Issue #9, opened 2022-08-03
Set up a CD workflow in GitHub Actions


johngrimes:
We should set up two workflows within GitHub actions:

* a "test" workflow, which builds the application and runs any tests upon a push to any branch
* a "deploy" workflow, which deploys the application somewhere and only runs on the main branch, following the successful execution of the "test" workflow

This will assist with the development of the software, and the demonstration of the evolving solution to stakeholders.

johngrimes:
@heathfrankel I'm not sure how you are currently hosting the software, and how it is deployed there? 

I think it might be worth moving the deployment to CSIRO public cloud and add automation going forwards - this is something I can help with.

We would probably just need to get the new URL whitelisted with Brett's server, if we want to use that for demonstrations.

PR #10, opened 2022-08-04
Add test workflow

johngrimes:
This change adds a test workflow that:

- Builds the application
- Runs the tests
- Runs the linter

This partially addresses #9.


johngrimes:
I think we should merge this even though it is not passing yet, so that we have visibility of the status of our branches and proposed changes.

PR #14, opened 2022-08-08
Fix answerOption valueString and valueInteger not working

### LINKED ISSUE #4: answerOption.valueString not working
If answerOption.valueString is used the app truncates any subsequent form content.

This answerOption type is helpful when including answers that don't have an existing concept in a well known code system. They may be migrated to valueCoding eventually but this is helpful in the interim.

yeexianfong:
Resolves #4.


yeexianfong:
Ready for review :D

PR #15, opened 2022-08-12
Bug enablewhen answerboolean issue

### LINKED ISSUE #5: enableWhen.answerBoolean not working
When answerBoolean conditional beahviour is included, the questions will not display even if the condition is satisfied.

yeexianfong:
Resolves #5.

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -15,7 +15,6 @@ import { HttpClientModule } from "@angular/common/http";
 import {
   QuestionnaireItemStringComponent,
   QuestionnaireItemTextComponent,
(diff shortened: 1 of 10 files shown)


yeexianfong:
A potential issue is that if the enableWhen booleanValue: false, the questionnaire item will not be rendered at the start. 
It is supposed to be rendered, as all checkboxes are unchecked by default. It is included in backlog as #16.

It doesn't seem to be too closely related to this issue, so we can potentially merge this and solve #16 in its own PR.

yeexianfong:
I think it ready for review!

Issue #16, opened 2022-08-12
EnableWhen doesnt work immediately on initial form render


yeexianfong:

yeexianfong:
A way to fix this is to load questionnaire responses for linked checkbox questions to be loaded with default values i.e. false. 
Not sure if this seems like a hacky approach...

yeexianfong:
Ended up setting default value to false to ALL QuestionnaireItemBooleanComponents

PR #18, opened 2022-08-15
Fix render enableWhen questions when rendering form for the first time

yeexianfong:
Very simple fix that managed to get the job done, but a huge learning experience - understood how form values are converted into QuestionnaireResponses now.


yeexianfong:
Resolves #16.

PR #151, opened 2023-02-09
Fix assemble operation

yeexianfong:
Bunch of fixes in sdc-assemble and assemble-express


yeexianfong:
Resolves https://jira.csiro.au/browse/SMARTHC-231 and https://jira.csiro.au/browse/SMARTHC-232.

PR #187, opened 2023-03-27
Fix text calculated expressions

yeexianfong:


yeexianfong:
Refactor components and fix calculated expressions

Issue #508, opened 2023-10-17
Issue with Display Conditions in the Viewer


LarsDT:
Description:

We are working on creating medical histories for the medical context. To ensure clarity, we have defined display conditions within the questionnaire. Unfortunately, we have noticed that these display conditions are not being considered in the viewer.

Current State:
Currently, the viewer does not display any elements that are associated with display conditions. This means that essential questions and information are not visible, significantly affecting our work within the medical context.

Desired State:
The viewer should display the medical history in accordance with the defined display conditions. All elements that have display conditions should be shown or hidden as per the specified criteria. This is crucial to ensure targeted and organized medical history collection within the medical field.

We urgently need assistance to resolve this issue and ensure that the display conditions in the viewer are properly taken into account. We sincerely thank you in advance for your help and support.

fongsean:
Hi Lars, thanks for raising this. 

If possible, can you provide me a snippet of your questionnaire to ease testing things out?

LarsDT:
Hello, I apologize for not getting back to you sooner. We've already identified and resolved the issue. Thank you very much for your prompt response.

fongsean:
Sorry, linked the wrong issue. 

Thanks for the heads up! Will be closing this issue in that case.

PR #537, opened 2023-10-30
Date time to date conversion and Slider display value

fongsean:


fongsean:
Resolves #508 (or at least attempts to).

Adds a non-editable display showing the value of the slider. When slider hasn't been interacted with, value is shown as "-" indicating that no answer has been provided. It provides more visibility to the user that the slider item has a concrete answer when interacted.

PR #575, opened 2023-11-23
Feature/minimal app

fongsean:


fongsean:
Demo app not complete. Plan to come back to it next week

Issue #622, opened 2024-02-07
Response validity feedback from renderer


nielmdrec:
When we receive the QuestionnaireResponse from the renderer (e.g. via getResponse()), it would be quite helpful to  know if the response form is **valid** in terms of having answers for each of the questions (items) which are **required** in the Questionnaire.

For example, if our Questionnaire has 5 questions and 4 of those questions have item.required = true,
we would like to know if those 4 required questions (at least) are answered before saving the QuestionnaireResponse to our database.

In theory we could do this by adding our own variable in the Questionnaire and setting it to 'true' ourselves when all required questions are answered.  However this would mean each of our Questionnaires needs to support this variable (and it becomes a proprietary feature for our Questionnaires only).

fongsean:
Hi Niel, that's definitely a good feature to have and honestly should have been included early.
I'll start working on it and hope it can be delivered before the end of the week.

nielmdrec:
Thanks, Sean.
We look forward to having this feature.
Glad you think it would be useful too.

nielmdrec:
BTW, Sean, for this feature would it be possible to have the mandatory questions marked by say an asterisk (*)?
(Or should I raise that in a separate item?)

fongsean:
Hi Niel, that's definitely a part of my consideration! I'm trying to think of a best way to present visual cues. No need to raise in a separate item.

To extend my thoughts further.... instead of retrieving the "validated" status via getResponse(), I think it will be more robust if it an be exposed via the state event listeners i.e. ```const isValid = useQuestionnaireResponseStore.use.isValid()```. That way, it won't be tied to a button click.

I might also expose a listener to see what linkIds are failing validation if that is beneficial. Might also help verify if the required check is working properly on all items (it can get a bit complicated when working with repeat groups with further nested items)

fongsean:
Feel free to refer to https://github.com/aehrc/smart-forms/pull/634#issue-2141636314 for changes up until now.
It's deployed as @aehrc/smart-forms-renderer v0.13.1.

Any thoughts on the bugs + future tasks?

adha-admin:
Hi Sean, 
Been chatting with Brian and he suggests you return the linkIds are failing validation using an OperationOutcome Resource.
Code example here:

https://github.com/brianpos/fhir-net-web-api/blob/develop/src/Hl7.Fhir.StructuredDataCapture/QuestionnaireResponseValidator.cs

fongsean:
Hi Phil,

[...]

PR #634, opened 2024-02-19
Issue/622

fongsean:
Changes:
- Added "required" and "read-only" labels to items
- Enforce validation for minLength and maxLength
- Make validation feedback for minLength and maxLength more intuitive
- Expose **invalidItems** and **responseIsValid** hooks from questionnaireResponseStore

Known bugs:
- Upon initialisation of form, invalidItems is set to 0 and responseIsValid is set to true. They will only become accurate when the form is edited. Will fix soon.
- For required boolean items, not interacting with the checkbox will result in no answer (as intended). However, if the intended answer is false, user would have to check and uncheck the checkbox to register an answer.
- For required items with calculated expressions, if the calculated value is 0, the value doesn't change and its answer is therefore not registered. Eager to fix but don't see an obvious solution.

Future tasks:
- Would be nice to have a validateQuestionnaire() function to trigger invalid fields to be highlighted as red.


fongsean:
Resolves #622.

PR #636, opened 2024-02-20
Update data backups

### LINKED ISSUE #635: Back up resources from EHR and forms server
Our EHR and forms server deployments are hosted on AWS, which restarts itself every 45 or 60 days(?).
The deployments are using an in-memory database, so they get wiped when the server restarts.

There is a need to have data redundancy in a form of backups in git.

fongsean:
Resolves #635.


fongsean:
Backups are as of 20 Feb 2024.

PR #680, opened 2024-03-26
Feature/data-linkid attribute for event listeners

fongsean:
Added `data-linkid` attribute to most Questionnaire item containers.

For non-group item types:
- `data-linkid` attribute is present on standalone items

For group item types:
- `data-linkid` attribute is present in tables like grid table and group table. Attribute is present on the table as whole - items within will not have the attribute.
- `data-linkid` attribute is not present on groups and repeat groups. Instead, they are present on the items contained within these groups.


fongsean:
Changes are live as of v0.16.0 of https://www.npmjs.com/package/@aehrc/smart-forms-renderer

Issue #724, opened 2024-04-09
Support for maxDecimalPlaces validation


fongsean:
Reference:
https://build.fhir.org/ig/HL7/sdc/behavior.html
http://hl7.org/fhir/R4/extension-maxdecimalplaces.html

Originally raised in https://github.com/aehrc/smart-forms/issues/622#issuecomment-2044185759

fongsean:
Fixed in PR #744

PR #744, opened 2024-04-21
Issue/722

### LINKED ISSUE #722: Expose validations as OperationOutcomes
Items within the QuestionnaireResponse that are failing validation should be exposed as OperationOutcome Resources instead of just the questionnaire items. This makes it align more with FHIR and provide standard warning/error messages.

A reference implementation:
https://github.com/brianpos/fhir-net-web-api/blob/develop/src/Hl7.Fhir.StructuredDataCapture/QuestionnaireResponseValidator.cs

Originally raised from https://github.com/aehrc/smart-forms/issues/622#issuecomment-2044150599

fongsean:
Resolves #722.


fongsean:
Mistake - Resolves #724 instead of #722

Issue #737, opened 2024-04-17
Support the display of units of measure after the input field


adha-admin:
Example 

                {
                    "extension": [
                        {
                            "url": http://hl7.org/fhir/StructureDefinition/questionnaire-unit,
                            "valueCoding": {
                                "system": http://unitsofmeasure.org,
                                "code": "m",
                                "display": "meters"
                            }
                        }
                    ],
                    "linkId": "Item-decimal",
                    "text": "decimal",
                    "type": "decimal"
                },

PR #746, opened 2024-04-24
Issue/737

fongsean:
Added support for:
- minValue and maxValue validation
- questionnaire-unit display


fongsean:
Resolves #737

PR #867, opened 2024-06-14
Fix/re pop encounter

fongsean:


fongsean:
Resolves https://github.com/aehrc/smart-forms-facade/issues/28 for bringing in "deletions".

PR #987, opened 2024-09-02
Change most instances of cloneDeep to use immer's produce

fongsean:


fongsean:
Decided to use `structuredClone()` to replace `cloneDeep()` as per PR #1023.

Advantages:
1. `structuredClone()` is supported natively by all modern browsers, eliminating the need to bring it an additional dependency.
2. Immer makes the resulting object readOnly, which means it cannot be used properly when adding it to an object that requires mutation downstream i.e. questionnaireResponse item which needs to have its value changed

Issue #1007, opened 2024-09-17
Renderer's theme overriding app's styles


fongsean:
It has been reported that when an app is consuming the Smart Forms renderer library, the background colours in the parent app is overwritten.

Might be something to do with Material UI's themes which enforces styles.

fongsean:
Seems like `<CssBaseline />` is the culprit.

Apparently it provides baseline MUI styles for the app, which seems to make sense that it overrides the styles defined in the app previously. For more reading: https://mui.com/material-ui/react-css-baseline/

Fixed it by removing `<CssBaseline />` from `RendererThemeProvider`.

PR #1019, opened 2024-09-20
Add function to remove repeat IDs, fix an issue where repeating group…

fongsean:
…s will refresh instances due to the mismatch of internal IDs


fongsean:
This PR resolves #1007.

PR #1023, opened 2024-09-25
Fix/use structuredclone

fongsean:


fongsean:
### Changed
- Replaced all instances of cloneDeep() with the native structuredClone() function.
- Changed Storybook iframes in the documentation to be dynamically sized based on the content.
- Updated renderer version to 0.42.0

PR #1032, opened 2024-10-08
Fix/iframe resizer dependency

fongsean:


fongsean:
Merging this into main.

- Renderer library updated
- Storybook updated
- Documentation updated

PR #1039, opened 2024-10-17
Fix checkbox open label string input

### LINKED ISSUE #1036: Unexpected behaviour of open-choice 'other' option when using tabs
**Describe the bug**
Within a Questionnaire organised by tabs, and with a Questionnaire item of type `open-choice`, with either repeats = true or false, the free text entry box for 'other' is populated by the value of one of the choice options after first navigating to a different tab and going back to the tab with the open choice question.

**To Reproduce**
This is evident in the Aboriginal Torres Strait Islander Health Check questionnaire found on https://www.smartforms.io/standalone 

1. open the Aboriginal Torres Strait Islander Health Check form
2. go to the tab 'Ear health and hearing'
3. under otoscopy findings for either left ear or right ear, check one or more of the choices, leaving 'other, please specify' empty and unchecked; eg:
image

5. go to another tab, like 'mood' or 'gambling' etc
6. go back to the tab 'Ear health and hearing'
7. notice that the check box 'other' is now checked and that the entry box is prepopulated with one of the selections made in step 3; eg:
image

**Expected behavior**
When navigating back to the open-choice question from another tab, its state regarding previously answered options, should be retained. And that the 'other' option should not be checked/prepopulated

fongsean:
Resolves #1036.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ This log documents significant changes for the [@aehrc/smart-forms-renderer](htt
 
 Changelog only includes changes from version 0.36.0 onwards.
 
+## [0.44.2] - 2024-10-18
+### Fixed
+- Fixed an issue where inputs in checkbox open-choice's open label field does not update the QuestionnaireResponse. 
 
 ## [0.44.1] - 2024-10-14
 ### Fixed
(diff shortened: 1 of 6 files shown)


fongsean:
Commit [2d66fcc](https://github.com/aehrc/smart-forms/pull/1039/commits/2d66fccc0469ff0dc4149ee35e1d5591139224ed) should be called "Update package to v0.44.2".

Issue #1089, opened 2025-01-29
Collapsible extension not working on gtable group item


liambarnes:
**Describe the bug**
The [collapsible extension](https://build.fhir.org/ig/HL7/sdc/StructureDefinition-sdc-questionnaire-collapsible.html) does not appear to be working for gtable items (possibly other tables too) as it does for other group items.
https://smartforms.csiro.au/docs/sdc/advanced/question#collapsible

**To Reproduce**
Load attached questionnaire in playground.
[Questionnaire-ChronicConditionManagementPlan.json](https://github.com/user-attachments/files/18581806/Questionnaire-ChronicConditionManagementPlan.json)

**Expected behavior**
Only table item text displayed when collapsed similar to the adverse reaction risk group in the attached questionnaire.

Workaround is to add a collapsible group item wrapper for the table.

fongsean:
Resolved as of v1.0.0-alpha.36.

Working on https://alpha.smartforms.io/playground.

Thanks @janadh for the implementation!

PR #1118, opened 2025-03-06
- Possible fix for #1089. Accordion is now being wrapped with the Gro…

janadh:
…upTableView if the qItem has collapsible extension.


fongsean:
Hi @janadh I have merged it in my IDE, push a new package version and closed the issue - https://github.com/aehrc/smart-forms/issues/1089.

Thanks for the implementation!

Issue #1179, opened 2025-04-09
React 19 compatibility and migrate MUI to v7 + React Query to v5


fongsean:
Doing some work to make this work with React 17, 18 and 19 versions.

Might involve updating to the latest MUI version, including mui/core, mui/lab and mui/x-date-pickers.

It should also support NextJS apps.

fongsean:
We will be supporting this migration https://github.com/brianpos/fhirpath-lab/tree/develop-upgrade-csiro on fhirpath-lab

fongsean:
FYI @brendan-cheong https://github.com/aehrc/smart-forms/pull/1182#issuecomment-2795771013

fongsean:
Resolved in https://github.com/aehrc/smart-forms/pull/1182.

PR #1182, opened 2025-04-10
Alpha 1179

fongsean:
PR for issue #1179.


fongsean:
# Changes

**1. Support for React 19**
- The renderer officially only supports React 18 and React 19 from now onwards.
- It might work on React 17, but you need need to do `npm i --legacy-peer-deps` or `npm i --force`
- The renderer fully uses client-side rendering. If your app is running React 19 with server-side rendering, you might need to have "use client" at the top of the file that uses the <BaseRenderer/>. 

**2. Migrate MUI from v6 to v7 and React Query from v4 to v5**
- The main motivation behind this is support for React 19, which old versions of both packages do not fully support (due to their dependencies).

**3. Peer Dependency Changes**
- MUI and React Query peer dependencies are removed. The only peer dependencies left are for React (`react` and `react-dom`)
- The motivation behind this is that some notable adopters are not on React e.g. fhirpath-lab (Vue 2) and the ADHA's CHAP app (Angular), and therefore it does not make sense for them to have MUI and react-query - both are React libraries. Having these as peer dependencies also force React-based adopters to have MUI v7 and React Query v5 as their dependencies.
- MUI's https://mui.com/material-ui/react-grid/ is especially a painful component to deal with. In MUI v7, they have Grid and GridLegacy, but these used to be Grid2 and Grid respectively - this means theres no way for the renderer to support both MUI v6 and MUI v7, due to the usage of same names representing different components in these two versions.
- It still makes sense for `react` and `react-dom` to be peerDependencies to avoid duplication of react versions. A non-React app i.e. no `react` and `react-dom` as dependencies will always default to install the latest version of React 19.

**4. SDC-Populate and SDC-Assemble compiling to CJS and ESM**
- Previously, both packages only compile to CJS, which means they can't be used directly with ESM-based web apps in a Monorepo locally.
- Both packages now uses [tsup](https://github.com/egoist/tsup) which compiles to both CJS and ESM - apps will automatically pick the format they are compatible with. That means the [prior configs in VIte](https://vite.dev/guide/dep-pre-bundling.html#monorepos-and-linked-dependencies) are not required anymore.
- Reminder: To build SDC-Populate and SDC-Assemble locally, do not use `tsc` or `tsc -w`, use `npm run compile` and `npm run watch` respectively.

**5. SDC-Populate and SDC-Assemble Breaking Changes**
- There are some interfaces changes for `FetchResourceCallback` and `FetchTerminologyCallback` - around their request configs, FetchResourceRequestConfig and FetchTerminologyRequestConfig respectively.
- Changes including a standardisation of interface naming starting with `Fetch...`
- It is now a requirement for FetchResourceRequestConfig to have a `sourceServerUrl` as the FHIR server URL to fetch resources from.
- It is now a requirement for FetchTerminologyRequestConfig to have a `terminologyServerUrl` as the FHIR Terminology server URL to fetch terminology from.

PR #1201, opened 2025-04-15
Alpha mm template

MaryamMehd:
Initial implementation of template-based extraction.


packages/smart-forms-renderer/src/stores/selector.ts
fongsean, line 1: Not sure why this was changed - it looks ok to me but I'm hesitant to touch it - might break the Vue and Angular apps using the renderer

packages/smart-forms-renderer/src/utils/extractTemplate.ts
```diff
@@ -0,0 +1,1032 @@
...
+  }, {} as Record<string, any>);
+
+  // Add relevant assumptions about the mapping based on found patterns
+  if (Object.keys(debugInfo.fieldMapping.mappedFields).length === 0) {
+    debugInfo.fieldMapping.assumptions.push('No fields could be mapped between templates and questionnaire items');
+  } else {
+    debugInfo.fieldMapping.assumptions.push(`Successfully mapped ${Object.keys(debugInfo.fieldMapping.mappedFields).length} fields`);
+  }
+
+  // Check for specific patterns in the template (like height, weight, blood pressure)
```
fongsean, line 549: This looks like its just for debugging (correct me if i'm wrong) - in that case that's fine.
Eventually when migrating this to the sdc-template-extract, I think we can omit this part.

It is still super useful to keep debug information in sdc-template-extract, but it will be good to keep it as generic as possible

fongsean: This comment is for L549 onwards, and really all the BP/BMI specific stuff throughout the utils file

fongsean:
Looks good overall, I think the business logic is ready to be moved to a package. I will create one for this.

Issue #1186, opened 2025-04-11
Request support for answerOptionToggleExpression


nielmdrec:
**Is your feature request related to a problem? Please describe.**
We have 2 related questions (Type and Hallucination) and both are choice items with contained valuesets.
We want to limit the choices for Hallucination question based on the answer for Type.

**Describe the solution you'd like**
We want to use the answerOptionToggleExpression to implement the following behaviour:
In this scenario:
 * question **Type** has the following options: _None, Visual, Tactile, Auditory_
 * question **Hallucinations** has the following options: Lucid, Infrequent, Brief, Frequent, No meaningful reality
1. if Type answer is 'None', then we want to restrict the options for Hallucinations to just 'Lucid'
    (basically disabling all the other options)
2. if Type  answer is anything other than 'None', then we want to restrict the options for Hallucinations to:
   Infrequent, Brief, Frequent, No meaningful reality
   (basically disabling 'Lucid')
****
**Describe alternatives you've considered**
We have tried using initial to default the values or calculatedExpression to default the Hallucination question, but the behaviour is lacking.

**Additional context**
We would want it to work for contained Valuesets, not just Valuesets retrieved from ontoserver.

fongsean:
Hi @nielmdrec which version of the renderer is Medirecords using? 0.45.x? or alpha?

Support for answerOptionToggleExpression is now available at https://alpha.smartforms.io/playground (but not the main version). Might need some work porting over the implementation into main if that is what you are using.

Use the questionnaire here. Let me know how it goes.
```json
{
  "resourceType": "Questionnaire",
  "id": "AnswerOptionToggleExpressionContained",
  "name": "AnswerOptionToggleExpressionContained",
  "title": "AnswerOptionToggleExpressionContained",
  "version": "0.1.0",
  "status": "draft",
  "publisher": "AEHRC CSIRO",
  "date": "2024-05-01",
  "url": "https://smartforms.csiro.au/docs/behavior/choice-restrictions/answer-option-toggle-expression-2",
  "contained": [
    {
      "resourceType": "ValueSet",
      "id": "HallucinationType",
      "url": "https://smartforms.csiro.au/ValueSet/HallucinationType",
      "name": "HallucinationType",
      "title": "Type of Hallucination",
      "status": "draft",
      "description": "Options for type of hallucinations",
      "compose": {
        "include": [
          {
            "system": "https://smartforms.csiro.au/CodeSystem/hallucination-type",
            "concept": [
              {
[...]

PR #1218, opened 2025-04-28
Alpha 1186

fongsean:
For issue #1186


fongsean:
Need to wait for alpha.48 to be published before publishing these changes

Issue #1219, opened 2025-04-28
Inline validation messages on field lose focus event


preeti-niranjane:
**Is your feature request related to a problem? Please describe.**
The inline validation messages are displayed as the user types in the field.

**Describe the solution you'd like**
The request is to provide ability to display the field validation messages on the lose focus event of any fields.

CC-

robeastwood-agency:
#### Further detail re technical background:

Using the extension targetConstraint (as implemented via #1066) to control input into a given form field eg a string item, with a constraint of regex to force exactly 10 digits. See simple Questionnaire below that demonstrates the issue.

**Current behaviour** is that, as the user starts entering digits, the warning is eagerly presented as they go, until the full 10 digits are entered. 

![Image](https://github.com/user-attachments/assets/c041ae80-6987-44e8-ba72-a1cab3646de4)

The **desired behaviour** is to render the validation message when the field loses focus instead of while it has focus and is still receiving input.

##### Sample Questionnaire

```
{
  "resourceType": "Questionnaire",
  "id": "example-questionnaire",
  "status": "active",
  "subjectType": [
    "Patient"
  ],
  "item": [
    {
      "linkId": "1.1",
      "text": "Enter 10 digit number",
      "type": "string"
    }
  ],
  "extension": [
    {
      "url": "http://hl7.org/fhir/StructureDefinition/targetConstraint",
      "extension": [
        {
          "url": "key",
          "valueId": "constraint"
        },
        {
          "url": "severity",
          "valueCode": "warning"
        },
        {
          "url": "expression",
          "valueExpression": {
            "language": "text/fhirpath",
            "expression": "%resource.item.where(linkId='1.1').answer.value.matches('^[0-9]{10}$').not()"
          }
        },
        {
          "url": "human",
          "valueString": "Number must be exactly 10 digits"
        },
        {
          "url": "location",
          "valueString": "Questionnaire.item[0]"
        }
      ]
    }
  ]
}
```

PR #1233, opened 2025-05-05
- Possible fix for Inline validation messages on field lose focus eve…

janadh:
…nt #1219

- We added event handlers for focus lost event (onBlur) and display the error message based on the state variables.


fongsean:
I've added the following changes:
- Abstracted showFeedback state to a custom hook
- Always show feedback after first blur

fongsean:
This change should be in version alpha.57

Issue #1144, opened 2025-03-26
Creating a prototype Template-based extraction algorithm


fongsean:
There are two examples that we can test against:

### extract-complex-template
```
{
  "resourceType": "Questionnaire",
  "id": "extract-complex-template",
  "meta": {
    "profile": [
      "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-extr-template"
    ]
  },
  "contained": [
    {
      "resourceType": "Patient",
      "id": "patTemplate",
      "identifier": [
        {
          "extension": [
            {
              "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractContext",
              "valueString": "item.where(linkId = 'ihi').answer.value"
            }
          ],
          "type": {
            "text": "National Identifier (IHI)"
          },
          "system": "http://example.org/nhio",
          "_value": {
            "extension": [
              {
                "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractValue",
                "valueString": "first()"
              }
            ]
          }
        }
      ],
      "name": [
        {
          "extension": [
            {
              "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractContext",
              "valueString": "item.where(linkId = 'name')"
            }
          ],
          "_text": {
            "extension": [
              {
                "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractValue",
                "valueString": "item.where(linkId='given' or linkId='family').answer.value.join(' ')"
              }
            ]
          },
          "_family": {
            "extension": [
              {
                "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractValue",
                "valueString": "item.where(linkId = 'family').answer.value.first()"
              }
            ]
          },
          "_given": [
            {
              "extension": [
                {
                  "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractValue",
                  "valueString": "item.where(linkId = 'given').answer.value"
                }
              ]
            }
          ]
        }
      ],
      "telecom": [
        {
          "extension": [
            {
              "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractContext",
[...]

PR #1256, opened 2025-05-22
Issue/1144

fongsean:
Initial working implementation for issue #1144.

Tested on https://build.fhir.org/ig/HL7/sdc/Questionnaire-extract-complex-template.html.

Merging into `alpha` before further iterations as it does not contains any notable changes in the renderer.


fongsean:
This PR also includes a change for the sdc-populate library - need to bump version to 4.1.0 and renderer version to 1.0.0-alpha.55. It now uses the same removeEmptyAnswers() function as the renderer, leveraging mapQItemsIndex() and getQrItemsIndex().

Issue #1299, opened 2025-06-09
Observation-based extract - invalid FHIR being created re author/performer


robeastwood-agency:
**Describe the bug**
For the CHAP form we are attempting to use Observation-based extract and have configured the form item (eg body weight measurement) to have the requisite configuration (Obs extract SDC extension, code, category). The CHAP app has a working extract using the CSIRO method and Observation instances are successfully created - however when validating against the relevant AU Core profiles and resource, there is an error re `Observation.author` as no such element exists - it should be `Observation.performer`

The generated Observation:
```
{
    "resourceType": "Observation",
    "status": "final",
    "id": "obs-3-HeightWeightResults-Weight",
    "code": {
        "coding": [
            {
                "system": "http://loinc.org",
                "code": "29463-7",
                "display": "Body weight"
            },
            {
                "system": "http://snomed.info/sct",
                "code": "27113001",
                "display": "Body weight"
            }
        ]
    },
    "subject": {
        "reference": "Patient/123456-re",
        "type": "Patient",
        "display": "Rob Eastwood"
    },
    "derivedFrom": [
        {
            "reference": "QuestionnaireResponse/3858a34f-42b0-4628-bf54-50a0d7d94555"
        }
    ],
    "effectiveDateTime": "2025-06-09T14:26:05+10:00",
    "issued": "2025-06-09T14:26:05+10:00",
    "author": {
        "reference": "Practitioner/12345",
        "type": "Practitioner",
        "display": "Jane Roberts"
    },
    "valueQuantity": {
        "value": 76,
        "unit": "kg",
        "system": "http://unitsofmeasure.org",
        "code": "kg"
    },
    "category": [
        {
            "coding": [
                {
                    "system": "http://terminology.hl7.org/CodeSystem/observation-category",
                    "code": "vital-signs",
                    "display": "Vital Signs"
                }
            ]
        }
    ]
}
```

The `author` node needs to instead be as follows, with an array, instead of object, in order to pass validation:
```
"performer": [
        {
            "reference": "Practitioner/12345",
            "type": "Practitioner",
            "display": "Jane Roberts"
        }
],
```

**To Reproduce**
The relevant Questionnaire.item
```
{
                            "extension": [
                                {
                                    "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-observation-extract-category",
                                    "valueCodeableConcept": {
                                        "coding": [
                                            {
[...]

Issue #1290, opened 2025-06-05
Empty String passing validation on questionnaire


jdksloan:
**Describe the bug**
When using a text field, if you put a space in the box it will return that the responseIsValid

**To Reproduce**
Create a questionnaire with a text field and make it required. Add a space to the generated field and check the responseIsValid property

**Expected behavior**
I expect that this will not pass as it's just an empty string/white space.

I have looked into the code and found things like:
removeEmptyAnswersFromResponse

But that also doesn't seem to work, as it returns a new response and my original response is still valid.

fongsean:
Hi @jdksloan it should be fixed in 1.0.0-alpha.68. Thanks for reporting the issue!

PR #1320, opened 2025-06-16
Fix/issue 1299 maryam

MaryamMehd:
Fix issue #1299: Improve observation extraction from QuestionnaireResponse
Remove problematic removeIfNull calls, add comprehensive field mapping, enhance status handling


fongsean:
Resolves:
- #1290
- #1299

PR #1425, opened 2025-07-24
fix: sync inline validation messages with debounced QuestionnaireResp…

### LINKED ISSUE #1409: Sync inline validation messages with debounced QuestionnaireResponse updates
There’s a noticeable timing mismatch between inline validation messages and the validation summary banner.

This happens because:
- Text-based inputs use a 300ms debounce before updating the QuestionnaireResponse, to avoid performance issues on every keystroke.
- Inline error messages are shown immediately based on local input state.
- Validation summary banner is based on the updated (debounced) QuestionnaireResponse.

As a result, error messages may appear/disappear at different times across the UI.

We should explore aligning the inline validation display timing with the QR update, so validation messages and the banner update in sync. This might require some tweaking around the debounce duration.

MaryamMehd:
…onse updates

Resolves #1409

## Problem
There was a timing mismatch between different validation UI elements:
- Inline error messages appeared immediately when users typed invalid input
- Validation summary (JSON panel) updated only after 300ms debounce delay
- This created a jarring UX where validation states appeared/disappeared at different times

## Solution
1. Created useSynchronizedValidation hook that:
   - Maintains original immediate validation feedback behavior
   - Adds synchronization logic to align validation states
   - Provides manual sync capability on blur events

2. Enhanced StringItem component validation UX:
   - Validation errors now appear immediately on first keystroke
   - Errors remain visible while typing (better UX)
   - Added useEffect to automatically show feedback when errors are present

## Key Changes
- Added useSynchronizedValidation.ts hook for validation timing synchronization
- Updated StringItem.tsx to use synchronized validation and improved error display logic
- Maintains 300ms debounce for performance while fixing timing inconsistencies

## Testing
- Verified with validation tester in Storybook
- Tested with temporary 2-second debounce to clearly observe timing improvements
- Confirmed immediate error visibility and better synchronization

### DIFF
--- a/packages/smart-forms-renderer/src/components/FormComponents/StringItem/StringItem.tsx
+++ b/packages/smart-forms-renderer/src/components/FormComponents/StringItem/StringItem.tsx
@@ -15,43 +15,24 @@
  * limitations under the License.
  */
 
-import React, { useCallback, useState } from 'react';
-import type {
-  PropsWithFeedbackFromParentAttribute,
-  PropsWithIsRepeatedAttribute,
-  PropsWithIsTabledRequiredAttribute,
-  PropsWithItemPathAttribute,
-  PropsWithParentIsReadOnlyAttribute,
-  PropsWithParentStylesAttribute,
-  PropsWithQrItemChangeHandler,
-  PropsWithRenderingExtensionsAttribute
-} from '../../../interfaces/renderProps.interface';
-import type { QuestionnaireItem, QuestionnaireResponseItem } from 'fhir/r4';
-import useValidationFeedback from '../../../hooks/useValidationFeedback';
+import React, { useCallback, useState, useEffect } from 'react';
+import type { BaseItemProps } from '../../../interfaces/renderProps.interface';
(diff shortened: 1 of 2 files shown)


fongsean:
@MARYAMMEHDIZ I had a quick look and there's too many useEffects in there. Effects are mainly used to sync local state with external state, which I think in this case usage of effects can be avoided. The other reason is code with effects will be harder to maintain in the future, since it's very hard to pinpoint the sequence in which those effects are executed.

Few things:
1. Have a read through https://react.dev/learn/you-might-not-need-an-effect.
2. I think you should be able to use the `useDebounceValue` hook (from usehooks-ts) to "delay" the feedback from showing up. That way it avoids you having to use effects.
3. Have you configured the Prettier VSCode extension? There is also an "enable Prettier on save" setting you can enable. That way you don't have to run `npm run lint-fix` and `npm run prettier-fix` everytime.

fongsean:
Closing this as issue has been resolved in https://github.com/aehrc/smart-forms/pull/1430.

Issue #1416, opened 2025-07-22
Column width format issues with hidden items/columns


liambarnes:
**Describe the bug**
The attached Questionnaire has hidden items that correspond to columns 2 & 3. When the screen width is narrow, it seems fine, but as it gets bigger the blank space becomes larger. See pic.

<img width="1598" height="1216" alt="Image" src="https://github.com/user-attachments/assets/4dd48693-a8cc-4627-b58d-f8c0a97df0ca" />

**To Reproduce**
Load attached Questionnaire into playground. Adjust screen width to view behaviour. 

**Expected behavior**
No blank column space should be present.

[Questionnaire-Examination(hidden-columns).json](https://github.com/user-attachments/files/21355659/Questionnaire-Examination.hidden-columns.json)

janadh:
Issue is caused when the columnHeader is calculated including the hidden items. So if the row has hidden items they will end up as an empty space column which in turn adds extra columns in the row. 

After implementing it looks like below. 

<img width="987" height="1291" alt="Image" src="https://github.com/user-attachments/assets/13f86513-aa62-4be6-bf81-413a8461cf00" />

fongsean:
Resolved in https://github.com/aehrc/smart-forms/pull/1426.

PR #1426, opened 2025-07-24
- Fix for #1416

janadh:
- Issue is caused when the columnHeader is calculated including the hidden items. So if the row has hidden items they will end up as an empty space column which in turn adds extra columns in the row.


fongsean:
@janadh See https://github.com/aehrc/smart-forms/blob/alpha/packages/smart-forms-renderer/src/components/FormComponents/Tables/GroupTable.tsx#L80-L97, should be able to use `isItemHidden` this instead

fongsean (approved):

Issue #1422, opened 2025-07-23
Invalid items not updated when target constraint unmet by item hidden by enableWhen


brendan-cheong:
Our testers have noticed that items that have unmet target constraints, which are then hidden by an enableWhen, still appear as an invalidItem.

**Repro steps:**

1.	With [ComprehensiveHealthAssessmentProgram-en-AU-v1.0.0.json](https://github.com/user-attachments/files/21097285/ComprehensiveHealthAssessmentProgram-en-AU-v1.0.0.json) in the Alpha Playground, select _Yes, enter new measurements_ (third radio option) for _1. Can you check the patient's height, weight and waist measurements?_ in the _3. Basic examinations_ tab group. Observe that the _Height_, _Weight_ and _Waist_ questions display and the _QR_ _InvalidItems_ are populated with the item.
2.	Enter _999_ for _Height_. Observe that the inline message _Check the value, it seems too high_ displays.
3.	Select _Unable to check_ (first radio option) for _1. Can you check the patient's height, weight and waist measurements?_. Observe that although the _Height_ item is no longer visible, the item appears still appears within the _QR_ _InvalidItems_.

**Expected behaviour:**
Items hidden by an enableWhen are excluded from validation and do not appear as an invalidItem.

fongsean:
Hi @brendan-cheong this is now resolved in alpha.85. (along with https://github.com/aehrc/smart-forms/issues/1409)

See [this PR comment](https://github.com/aehrc/smart-forms/pull/1430#issuecomment-3134691599) for changes related to validation.

There has been a bit of changes going on - I recommend you to refer back to the [changelog](https://github.com/aehrc/smart-forms/blob/alpha/CHANGELOG.alpha.md) and view whats changed.

Won't close this off yet, let me know once it's tested and working on your end.

brendan-cheong:
Hey @fongsean, I've given this one a test and it's working well. Thanks!

PR #1430, opened 2025-07-25
fix: sync validation timing using useDebounceValue

### LINKED ISSUE #1409: Sync inline validation messages with debounced QuestionnaireResponse updates

MaryamMehd:
- Use useDebounceValue from usehooks-ts to delay feedback disappearance
- Show validation errors immediately (preserves UX)
- Errors disappear in sync with store validation timing (300ms debounce)
- Simple, clean solution without complex useEffect synchronization
- Resolves timing mismatch between inline validation and JSON panel

Addresses the feedback to use useDebounceValue instead of complex effects. Fixes #1409


fongsean:
### Changes:

- Refactor QuestionnaireStore `itemTypes` to `itemMap` - holds a <linkId, QItem (without qitem.item)> - supports resolving of `#1422`

- Fix Playground State store viewer to correctly display falsy values

- Refactor `useValidationFeedback` to rely entirely on the QR store’s `invalidItems`, ensuring feedback only shows based on validation state and resolving timing mismatches between inline and summary feedback - resolves #1409

- For DateTimeItem and QuantityItem, fix an issue where feedback was always shown — now only displays when `showFeedback=true`

- When processing targetConstraints, add a step to filter out `invalidItems` whose qItem is hidden via questionnaire-hidden or enableWhen - resolves #1422.

PR #1503, opened 2025-08-22
Refactor FHIR patch structure by removing unnecessary 'name' (only add operation)

### LINKED ISSUE #1502: Extract PATCH `name`parameter
In our extract parameters resource, we include a parameter called "name" ([example here](https://github.com/aehrc/smart-forms-ig/blob/bf5e2ef73a40816960ce5aa563f3a947e2f61bd1/input/examples/Bundle-sdc-template-extract-928bbdd27d15.json#L121C29-L124C30)).

According to the [FHIRPath Patch Profile](https://build.fhir.org/fhirpath-patch.html), this appears to be only used with an add operation. Since we use replace, we should remove it.

Change:
Only have `name` in add operations

fongsean:
Resolves #1502.


fongsean:
No code changes performed as this change does not affect the underlying implementation. Updated a few test cases.

PR #1576, opened 2025-09-18
fix: Time control now produces correct FHIR format (HH:mm:ss)

### LINKED ISSUE #1532: Time control produces invalid FHIR format
- [ ] TimeBasic
https://github.com/beda-software/smart-forms/blob/f5429aa51f3ee8e2fa1cb88bbd9c8b34ce060d9b/packages/smart-forms-renderer/src/stories/itemTypes/Time.stories.tsx

When we set to our timeInput "11:00 am"  format - the result.valueTime we get back has "2025-09-04T11:00:00+03:00" format, but we expect  "11:00:00".I put a screenshot below.

Image

Expect: "11:00:00"

MaryamMehd:
- Fix TimeItem.tsx to format time output as HH:mm:ss instead of full datetime
- Fix time parsing to properly handle time-only strings with dayjs format
- Resolves issue where time input '11:00 am' was producing '2025-09-04T11:00:00+03:00'
- Now correctly produces '11:00:00' as expected by FHIR specification

Resolves #1532

### DIFF
--- a/packages/smart-forms-renderer/src/components/FormComponents/TimeItem/TimeItem.tsx
+++ b/packages/smart-forms-renderer/src/components/FormComponents/TimeItem/TimeItem.tsx
@@ -49,13 +49,13 @@ function TimeItem(props: BaseItemProps) {
   if (qrItem?.answer && qrItem?.answer[0].valueTime) {
     timeString = qrItem.answer[0].valueTime;
   }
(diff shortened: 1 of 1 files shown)


fongsean (approved):
LGTM. Thanks @MaryamMehd. I will bump version once these two PRs are merged.

Could you also run linting before merging?

Issue #1610, opened 2025-09-26
Disable Cache for Calculated Expressions


janadh:
Variables using FHIRPath does not calculcate the values due to caching. 

Proposed fix: Disable cache on calculated expressions temporarily.

Issue #1611, opened 2025-09-26
Stale results from async expression cache


fongsean:
**Describe the bug**
The cache key for async calls is too aggressive. When evaluating expressions, the cache does not take the previous value into account, which causes stale or incorrect results to be returned.

**To Reproduce**
1. Open a form with an async expression (e.g. memberOf() or terminology lookup).
2. Trigger the async evaluation more than once with different values.
3. Observe that the cached result is reused even when the input value has changed.

**Expected behavior**
The async evaluation should consider both the expression and the previous value when generating the cache key, so that results are correctly recomputed when the input changes.

fongsean:
Had a look at this and designing a proper caching mechanism for async fhirpath evals has proven tricky. See the example below:

```json
{
  "url": "http://hl7.org/fhir/StructureDefinition/variable",
  "valueExpression": {
    "name": "memberOfBpLoweringMedicationCode",
    "language": "text/fhirpath",
[...]

PR #1619, opened 2025-10-01
- Temp fix for #1610 by disabling cache

janadh:


fongsean (approved):
LGTM. Given this is a temporary fix that will be resolved by https://github.com/aehrc/smart-forms/issues/1611, this change will not be published into the public renderer package.

AWS Amplify deployments on alpha.smartforms.io will still have this change.

Issue #1616, opened 2025-09-30
Quantity stories test bug


Vlpros:
In smart-forms-render/src/stories/itemTypes/Quantuty.stories.tsx  - We have 2 drop down tests.
- QuantityBasicComparator : after the input was filled we expect our comparator "<", but from store returns undefined 

<img width="576" height="741" alt="Image" src="https://github.com/user-attachments/assets/184b9b05-962d-4411-bd8a-fe207c53fb65" />

- QuantityMultiUnit : after the input was filled we expect out unit "Week(s)", but from store returns "Day(s)"

<img width="576" height="741" alt="Image" src="https://github.com/user-attachments/assets/b2f9d29f-4a0a-4201-aa01-11611d0d8e3d" />

fongsean:
Closing as the Quantity component does not actually contain any issues, the issue lies in the testing function.

See https://github.com/aehrc/smart-forms/commit/4cc0bcf28597f64322c438d3fd5343d5c24e6457#diff-a4351d9843202c6e77ac897a426badc4483a2e351be8262f9b9a50e1a0be4ffeR74-R88 for fix.

PR #1630, opened 2025-10-03
fix: resolve quantity stories bug #1616

MaryamMehd:
- Fix comparator selection being lost when selected before entering value
- Fix unit selection being lost when selected before entering value
- Always update QR item for comparator and unit changes, even with empty values
- Handle empty parsedNewInput in createQuantityItemAnswer to prevent NaN
- Add debug logging to track QR updates
- Add test stories QuantityBasicComparator and QuantityMultiUnit

Resolves: #1616


fongsean:
Closing as the Quantity component does not actually contain any issues, the issue lies in the testing function https://github.com/aehrc/smart-forms/blob/alpha/packages/smart-forms-renderer/src/stories/testUtils.ts#L46.

PR #1677, opened 2025-10-20
Evaluate all templateExtract extensions on a QItem

leoniedickson:


fongsean (changes requested):
LGTM!! It looks amazingly clean, esp with this part:
`.map((ext) => validateAndExtractTemplateExtractReference(ext, itemId))`

One more thing to do around version bumping:
1. Update sdc-template-extract/package.json version and its dependency version in smart-forms-app/package.json, then do a `npm i` to update package-lock.json.
2. Update sdc-template-extract/CHANGELOG.md. This file doesn't exist yet (at least in your branch), so OK to skip this step for this PR. I will do this step later.
3. Commit the 3 package.json/package-lock.json files.
4. Run `npm run build`.
5. Run `npm publish`. Might need to sort out NPM access if you don't have it already.

Edit:
Looks like you need to merge `alpha` into your branch because it's currently v1.0.13 - then update version to v1.0.14. You can also do step 2 if you like.

Issue #1641, opened 2025-10-09
Associate item measurement units (eg cm) to its parent item (accessibility)


robeastwood-agency:
**Is your feature request related to a problem? Please describe.**
For the purposes of meeting accessibility requirements, when an item has an associated measurement unit (eg cm or kg), the content is visibly associated to the on-screen field, but not so when a screen reader reads the content. The screen reader does not announce the associated units.

**Describe the solution you'd like**
From an accessibility perspective, it is preferrable to have the measurement unit programmatically associated with its parent field, so that the content is read out immediately when the field receives focus. Accessibility experts suggest the use of the `aria-describedby` attribute to associate unit text with the input field.

**Describe alternatives you've considered**
None

**Additional context**
None

PR #1699, opened 2025-10-22
feat: Add aria-label accessibility for unit fields (issue #1641)

MaryamMehd:
- Add aria-label to Decimal, Integer, and Quantity field inputs

- Unit is announced by screen readers (e.g., 'Height (cm)')

- Add Storybook accessibility tests for all three item types

- Clean implementation with 0 linter errors

Resolves: https://github.com/aehrc/smart-forms/issues/1641

### DIFF
--- a/packages/smart-forms-renderer/src/components/FormComponents/DecimalItem/DecimalField.tsx
+++ b/packages/smart-forms-renderer/src/components/FormComponents/DecimalItem/DecimalField.tsx
@@ -86,7 +86,7 @@ function DecimalField(props: DecimalFieldProps) {
         htmlInput: {
           inputMode: 'numeric',
           pattern: '[0-9]*',
-          'aria-label': ariaLabel
+          ...(ariaLabel && { 'aria-label': ariaLabel })
         },
         input: {
           readOnly: readOnly && readOnlyVisualStyle === 'readonly',
--- a/packages/smart-forms-renderer/src/components/FormComponents/IntegerItem/IntegerField.tsx
+++ b/packages/smart-forms-renderer/src/components/FormComponents/IntegerItem/IntegerField.tsx
@@ -87,7 +87,7 @@ function IntegerField(props: IntegerFieldProps) {
         htmlInput: {
           inputMode: 'numeric',
           pattern: '[0-9]*',
-          'aria-label': ariaLabel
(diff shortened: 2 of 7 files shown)


packages/smart-forms-renderer/src/components/FormComponents/DecimalItem/DecimalField.tsx
```diff
@@ -80,7 +85,8 @@ function DecimalField(props: DecimalFieldProps) {
       slotProps={{
         htmlInput: {
           inputMode: 'numeric',
-          pattern: '[0-9]*'
+          pattern: '[0-9]*',
+          'aria-label': ariaLabel
```
fongsean: This is a good change.

Our primary way of applying accessible labels is via `htmlFor` in `ItemLabel.tsx`, so a cleaner way to conditionally apply `aria-label` is to use a spread operator instead of assigning undefined to `aria-label`:

```ts
htmlInput: {
  inputMode: 'numeric',
  pattern: '[0-9]*'
  ...(ariaLabel && { 'aria-label': ariaLabel })    // Conditionally apply label with unit if a unit is provided
}
```

packages/smart-forms-renderer/src/components/FormComponents/IntegerItem/IntegerField.tsx
```diff
@@ -81,7 +86,8 @@ function IntegerField(props: IntegerFieldProps) {
       slotProps={{
         htmlInput: {
           inputMode: 'numeric',
-          pattern: '[0-9]*'
+          pattern: '[0-9]*',
+          'aria-label': ariaLabel
```
fongsean: Same change as above here.

packages/smart-forms-renderer/src/components/FormComponents/QuantityItem/QuantityField.tsx
```diff
@@ -76,7 +83,7 @@ function QuantityField(props: QuantityFieldProps) {
       isTabled={isTabled}
       size="small"
       slotProps={{
-        htmlInput: { inputMode: 'numeric', pattern: '[0-9]*' },
+        htmlInput: { inputMode: 'numeric', pattern: '[0-9]*', 'aria-label': ariaLabel },
```
fongsean: Same change as above here.

PR #1700, opened 2025-10-22
feat: Associate instruction text with input fields using aria-describ…

### LINKED ISSUE #1640: Associate item instruction content to its parent item (accessibility)
For the purposes of meeting accessibility requirements, when an item has instructional text, the content is visibly associated to the on-screen field, but not so when a screen reader reads the content. The screen reader user has to tab to invoke the instructional text.

Example item content 
```
"item": [
    {
        "extension": [
            {
                "url": "http://hl7.org/fhir/StructureDefinition/questionnaire-displayCategory",
                "valueCodeableConcept": {
                    "coding": [
                        {
                            "system": "http://hl7.org/fhir/questionnaire-display-category",
                            "code": "instructions"
                        }
                    ]
                }
            }
        ],
       "linkId": "1-Background-Chap1Tel-Instructions",
       "text": "Please enter a valid Australian phone number",
       "type": "display"
    }
]

```

**Describe the solution you'd like**
From an accessibility perspective, it is preferrable to have the instructional text programmatically associated with its parent field, so that the content is read out immediately when the field receives focus (rather than having it need a tab key press). Accessibility experts suggest the use of the `aria-describedby` attribute to associate hint text with the input field.

**Describe alternatives you've considered**
None

**Additional context**
None

MaryamMehd:
…edby (#1640)

This commit implements accessibility improvements by associating instructional text with their parent form fields using the aria-describedby attribute.

Changes:
- Modified ItemFieldGrid to recursively add aria-describedby to input/textarea elements and radio/checkbox groups when instructions are present
- Added getInstructionsId helper function to generate instruction IDs
- Updated DisplayInstructions component to accept an optional id prop
- Added Storybook tests for String, Boolean, and Text items to verify aria-describedby functionality with instruction text

Screen readers will now announce instructional text when a field receives focus, improving the accessibility experience for users with visual impairments.

Fixes #1640

### DIFF
--- a/packages/smart-forms-renderer/src/stories/itemTypes/Boolean.stories.tsx
+++ b/packages/smart-forms-renderer/src/stories/itemTypes/Boolean.stories.tsx
@@ -27,7 +27,7 @@ import {
   questionnaireFactory,
   questionnaireResponseFactory
 } from '../testUtils';
-import { expect, fireEvent } from 'storybook/test';
+import { expect, fireEvent, within } from 'storybook/test';
 import { createStory } from '../storybookWrappers/createStory';
 
 // More on how to set up stories at: https://storybook.js.org/docs/react/writing-stories/introduction#default-export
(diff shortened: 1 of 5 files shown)


packages/smart-forms-renderer/src/stories/itemTypes/Boolean.stories.tsx
fongsean: Similar to https://github.com/aehrc/smart-forms/pull/1699, can you add them under the testing folder instead? Something like title: 'Testing/Accessibility' is fine.

fongsean (changes requested):
I don't think this actually fixed the problem. I went into the stories and tried Voiceover - couldn't get the instructions to be read out.

Can you try using aria-describedby in the MUI components instead of getting React to create them dynamically?

PR #1708, opened 2025-10-23
Issues/#1505

clinnygee:


clinnygee:
Hey, still have some clean up to do but i wanted it to run through CI and check for any issues. I will return and finish cleaning things up if I know that the functionality is there.

packages/smart-forms-renderer/src/components/FormComponents/SingleItem/SingleItem.tsx
```diff
@@ -77,7 +77,21 @@ function SingleItem(props: SingleItemProps) {
         parentIsRepeatGroup ? (parentRepeatGroupIndex ?? null) : null
       );
 
-      if (qrItem && qrItem.item && qrItem.item.length > 0) {
+      if (qItem.repeats && qrItem?.answer && qrItem.answer.length > 0) {
```
fongsean: I think you can remove `qItem.repeats` here. Nested items should still work on non-repeating items.

Try it out with this questionnaire:
```
{
  "status": "draft",
  "item": [
    {
      "linkId": "parent-decimal",
      "type": "decimal",
      "text": "Parent decimal with initial",
      "initial": [
        {
          "valueDecimal": 1
        }
      ],
      "item": [
        {
          "linkId": "child-string",
          "type": "string",
          "text": "Child string with initial",
          "initial": [
            {
              "valueString": "child value"
            }
          ]
        }
      ]
    }
  ]
}
```

packages/smart-forms-renderer/src/utils/initialise.ts
```diff
@@ -168,7 +168,14 @@ function readInitialValuesRecursive(
         text: qItem.text
       };
     }
-    qrItem.item = initialValues;
+
+    if (qItem.repeats) {
```
fongsean: I think some changes need to be done here to support the above questionnaire as well.

Instead of `qItem.repeats` might need to check `qItem.item && qItem.type !== group` or some sort.

packages/smart-forms-renderer/src/utils/qrItem.ts
```diff
@@ -221,3 +235,80 @@ export function updateQrItemsInGroup(
     }
   }
 }
+
+/**
+ * Updates the QuestionnaireResponseItem group by adding/removing a new/modified child QuestionnaireResponseItem into/from a qrGroup
```
fongsean: Looks good! Might or might not need any changes here to support a non-repeating case - just flagging for now.

Issue #1722, opened 2025-10-24
Open-choice control does not load the value from a variable through calculated expressions (When terminology server is being used)


janadh:
Open-choice items were not correctly loading calculated values. 
This is because the terminology valueSets are not being cached due to its size, so the calculated expressions have no values to match with the given input that was provided through the FHIRPath variable. 
Proposed solution is to check if matching fails, and if it is a coding object,  assign the input variable to its current value.

janadh:
Fixed in renderer version v1.2.1 and above

Issue #1721, opened 2025-10-24
open-choice control not storing selected option in calculatedExpressions


janadh:
When the open-choice item is used and when the user selects an option from the autocomplete items, the control fires onchange and input change and always keeps the entered text value and not the selected option value. This is only applicable when calculated expressions are used to get the selected value from the control.

PR #1723, opened 2025-10-24
Fix for Open-Choice  input behavior and Calculated Expressions

janadh:
Refactor OpenChoiceAutocompleteItem to handle selectOption reason correctly and optimize calculated expression matching logic

- Fix for #1721 and #1722


fongsean (approved):
LGTM!

PR #1736, opened 2025-10-27
Issue/1682 Fix to ensure animation runs on first calculated expression

leoniedickson:


fongsean (approved):
LGTM! Tested it locally on my end as well.

PR #1793, opened 2025-12-11
Issue/1780

MaryamMehd:


LOCAL_DEVELOPMENT.md
```diff
@@ -111,6 +111,10 @@ The Smart Forms app uses a `config.json` file for configuration, fetched at runt
      //  "https://example.com/fhir": "6cc9bccb-3ae2-40d7-9660-22c99534520b"
      // }
      registeredClientIdsUrl: string | null;
+     
+     // (Optional) Feature flag to enable/disable in-app population functionality
+     // Defaults to true if not specified
+     inAppPopulate?: boolean;
```
leoniedickson: I did a search for inAppPopulate in the repo and it doesn't seem to be referred to anywhere - so maybe this variable doesn't need to be added? Correct me if I'm missing anything

MaryamMehd: Good catch! You're right - inAppPopulate was defined in the config but wasn't actually being used anywhere in the app to control any behavior.
I've removed it in commit 1e5f6c4eb. 

Note on CI: The Playwright test (saving.spec.ts) is failing due to timeouts with the external proxy.smartforms.io service - this is unrelated to the config changes. All other tests pass. Let me know if you'd like me to retry the CI or if this can be merged despite the flaky E2E test.

leoniedickson (approved):
Looks good!

PR #1818, opened 2026-02-17
Make changes around padding and margins to reduce vertical whitespace…

clinnygee:
… on the screen.


leoniedickson (commented):
It's definitely looking nicer! Was having a look with the GPChronicConditionManagementPlan form and noticed there were a few spots where the container borders are touching - maybe unintentionally? 

Image
Image


Also just had another thought that the add item buttons could possibly be moved to sit next to the remove icon on the last item to get rid of that vertical space where the button is now (bad mockup below) - just an idea though!

Image

Looks like prettier might also need to be run to get that check passing

PR #1814, opened 2026-02-06
Behaviour test. Aboriginal form condition logic

ruscoder:
Hello,
@roman-podrubny and I prepared a PR with behaviour tests for Aboriginal Health Check form for enableWhen expressions.

This PR includes:
- The snapshot of AboriginalTorresStraitIslanderHealthCheckAssembled-0.4.0 copied from https://smartforms.csiro.au/ with removed readOnly attributes in order to make it editable through these tests.
- [All Enable when expressions](apps/smart-forms-app/src/test/aboriginalFormConditions.test.tsx) (except age-related - they will be covered in the separate PR)
- Added missing data-label/data-test attributes in order to refer to specific form items
- Added addition helpers for interaction with the form (for checkbox, for tabs)

The tests are running using playwright with an existing command `npm run test-715` that runs only these specific test suites.
Because of the high number of tests, it takes at least ~15min to run 138 tests. So, maybe them should not be run on every commit, and instead, from time to time to make sure that there's no regression in the AboriginalTorresStraitIslanderHealthCheck form.

Currently we're working on:
- Covering all calculation logic
- Filling the form by different age groups
We will create PR as soon as it's ready

PR #1821, opened 2026-02-25
Cover aboriginal form calculatedExpression

roman-podrubny:
Hi, 
We covered aboriginal form all calculatedExpression logic.
While covered we found bug in questionnaire related to Allergies/adverse reactions Health priorities, actions and follow-up in summary. This test is marked as skipped. I'll create separate issue about that.

Please take a look this PR after #1814 is merged into main.


ruscoder:
@MaryamMehd HI, could you please do a review of this PR. Now, after merging the previous PR, it constains only few changes and test itself.

PR #1828, opened 2026-03-03
Issue/1824

leoniedickson:
Also fixes errors in ci tests


MaryamMehd (approved):
The calculatedExpression fix looks good, 
and the Jest config fix resolves the test failures from #1827.

PR #1830, opened 2026-03-05
Issue/1640

MaryamMehd:
## Testing Instructions

Hi @leoniedickson ! This PR rebases issue #1640 on the latest main and fixes the lint error that was blocking CI/CD.

### What Changed:
- Fixed lint error (removed unused `within` import)
- Rebased on latest main
- All tests passing

### How to Test:

cd packages/smart-forms-renderer
npm run storybook

#### Option 1: VoiceOver Test (Recommended)
1. Navigate to any **ItemType** story with instructions (e.g., String, Boolean, Integer)
2. Find a field with instruction text (gray text below input)
3. Press `⌘ Cmd + F5` to **activate VoiceOver**
4. Tab to the input field
5. **Listen:** VoiceOver should announce the instruction text automatically ✅
6. Press `⌘ Cmd + F5` to stop VoiceOver

#### Option 2: Inspect HTML Elements
1. Navigate to any **ItemType** story with instructions
2. Right-click on an input field → **Inspect**
3. Verify the `<input>` element has `aria-describedby="instruction-id-xxx"`
4. Verify the instruction `<span>` element has matching `id="instruction-id-xxx"`
5. Confirm they are correctly linked

### Manual Testing Completed:
- Tested with VoiceOver in both Storybook and main app
- Instructions are announced correctly when field receives focus
- All unit tests passing (598 tests)
- Lint checks passing (0 errors)

Ready for review!


leoniedickson (changes requested):
Hey @MaryamMehd, I tried out the voiceover stories in storybook and it's looking like a good start! I did notice that not all type items are covered - e.g. after prepopulating 715, the "About the health check" tab has the "Eligible for health check" choice radio group item which has instructions not included in the voiceover.

Looks like string, text, boolean are covered - the other item types are integer, decimal, quantity, url, attachment, date, dateTime, time, choice, open-choice. I assume display instructions can be associated with all of these?

Claude also picked up a few issues that would be good to address:

---

**Bug: `slotProps.input` puts `aria-describedby` on the wrong element in StringField and TextField**

`slotProps.input` targets the `OutlinedInput` root `<div>`, not the underlying `<input>`/`<textarea>`. Screen readers expect `aria-describedby` on the actual form control element. Should use `slotProps.htmlInput` instead:

```tsx
// Current (targets wrapper div — wrong):
slotProps={{
  input: {
    ...(instructionsId && { 'aria-describedby': instructionsId }),
  }
}}

// Fix (targets the actual <input>/<textarea>):
slotProps={{
  htmlInput: {
    ...(instructionsId && { 'aria-describedby': instructionsId }),
  }
}}
```

This is consistent with how `BooleanField` handles it — via `inputProps` on `ChoiceRadioSingle`, which correctly targets the underlying `<input type="radio">`.

---

**Bug: `StringInstructionsAccessibility` story queries for `textarea` but string fields render `<input>`**

```ts
// This will always return null for a string item:
const textarea = inputField?.querySelector('textarea');
```

Should be `input` (or `input, textarea` to be safe). The `TextInstructionsAccessibility` story using `textarea` is correct.

---

**Cleanup: `'aria-describedby'?: string` type addition in `Radio.styles.tsx` is unnecessary**

`aria-describedby` is passed to `StandardRadio` via `inputProps`, not as a direct prop, so this type addition on the styled component has no effect and can be reverted.

leoniedickson (changes requested):
Hey @MaryamMehd, I have a few comments:

1. I'm concerned about the custom fields implemented for choice and open-choice dropdowns which have switched out Autocomplete for Select components - this removes the functionality that autocomplete components have (such as typing to filter options), and the open-choice dropdowns also no longer support valueString and valueInteger answer options. 

    I think this change needs to be reverted. Did you try targeting `slotProps.htmlInput` as mentioned in my last comment?

2. I tried out the example in the 715 in playground that I mentioned before (after prepopulating 715, the "About the health check" tab has the "Eligible for health check" choice radio group item with instructions), but still can't see that the voiceover includes those instructions. Can you check that instructions for all of the choice/open-choice variants are covered (checkbox/radio/select/autocomplete, options/value set)?

3. I'm not as concerned about the change to use the custom time item, but since it's a behaviour change, it's probably worth checking that it's okay from a product/user perspective? Or alternatively, if there's another solution like targeting `slotProps.field` (mentioned above) that works without changing the component, that would probably be more in scope for this issue. 

    If not, and if there's no issue with switching to the custom time field, it would be good to clean up the codebase since TimeItem isn't reached anymore and has duplicated logic to CustomTimeItem. (There's also a bug in `convertToFhirTime` on line 135 as the parsing for 12 hour time should use 'hh' instead of 'HH')

4. I can't see the voiceover for Attachment instructions working in storybook (I don't know if this is an issue with the story or with the implementation in the component)

5. I think I also missed this one before, but we also have the `slider` item control which can have instructions

6. Looks like there are some merge conflicts :)

MaryamMehd:
Hey @leoniedickson , Thank you for your thorough comments, was very helpful.

I addressed all your comments, it is ready for your review now.

leoniedickson (approved):
Thanks @MaryamMehd, looks good to me!

PR #1859, opened 2026-03-25
temp

clinnygee:


leoniedickson (changes requested):
Hey @clinnygee, looks good but seems like there's one gap:

`RepeatItem/RemoveItemButton` still has `numOfRepeatAnswers === 1` in the disabled check, so the remove button remains disabled when there's only a single non-group repeat answer. The same fix applied to `RepeatGroup/RemoveItemButton` (dropping the count guard) should be applied here too — `RepeatItem.tsx` already handles the zero-answers case correctly (emits an empty qrItem when the last answer is removed).

clinnygee:
> Hey @clinnygee, looks good but seems like there's one gap:
> 
> `RepeatItem/RemoveItemButton` still has `numOfRepeatAnswers === 1` in the disabled check, so the remove button remains disabled when there's only a single non-group repeat answer. The same fix applied to `RepeatGroup/RemoveItemButton` (dropping the count guard) should be applied here too — `RepeatItem.tsx` already handles the zero-answers case correctly (emits an empty qrItem when the last answer is removed).

Hey @leoniedickson yes this looks like an oversight, but it behaves a bit weird like that because these fields already have their own clear button. I checked with Liam before changing this back to === 1. I attached a screenshot to
Screenshot 2026-04-20 at 10 48 31 am

leoniedickson (approved):

PR #1889, opened 2026-04-12
Issue/1794

### LINKED ISSUE #1794: Text styling on Questionnaire.title and Questionnaire.item.prefix
This ticket is a backlog item but not critical.

Text styling is implemented for item.text and answerOption.valueString. It is also valid to use on Questionnaire.title and Questionnaire.item.prefix. Please implement styling extensions to work on these elements. Extensions are:
- https://hl7.org/fhir/extensions/StructureDefinition-rendering-style.html
- https://hl7.org/fhir/extensions/StructureDefinition-rendering-markdown.html
- https://hl7.org/fhir/extensions/StructureDefinition-rendering-xhtml.html

This capability is specified in the [Advanced Rendering Questionnaire](https://build.fhir.org/ig/HL7/sdc/en/StructureDefinition-sdc-questionnaire-render.html) profile.

Relates to https://github.com/aehrc/smart-forms/issues/1786

MaryamMehd:


leoniedickson (changes requested):
Thanks @MaryamMehd looking good! Just some small bits of feedback that would be good to address:

**Title**

- **Playground title suppressed** — `resetAndBuildForm` always calls `setRendererConfig({ hideQuestionnaireTitle: true })`, so the playground (which has no app header rendering the title) also suppresses it. Suggest adding an optional `hideQuestionnaireTitle` parameter to `resetAndBuildForm` (defaulting to `false`) so the main app can pass `true` and the playground gets the title by default.

- **Title spacing** — the title renders tight against the top/left of the container in `BaseRenderer`. Needs some padding/margin.

- **`RendererHeader` should use plain text** — the app header should just render `sourceQuestionnaire.title` as plain text rather than using `<QuestionnaireTitleText>`.

- **Storybook double title** — `BuildFormWrapperForStorybook` renders `QuestionnaireTitleText` explicitly, and `<BaseRenderer />` now also renders it since `hideQuestionnaireTitle` defaults to `false`. Remove the manual title from the wrapper.

- **Minor: `_title`-only questionnaires** — the guard in `BaseRenderer` is `sourceQuestionnaire.title`, so a questionnaire with `_title` rendering extensions but no `title` string won't render the title at all. Should also check `sourceQuestionnaire._title`.

---

**Prefix**

- **Bug: `itemHasLabelHeadingContent()` too broad** — returns `true` for any extension on `_prefix`, not just rendering or calculated ones. A group with no `text`, no `prefix`, but an unrelated extension on `_prefix` (e.g. `translation`) will render an empty heading. Should check for specific rendering/calculated extension URLs only. The existing test for this function doesn't cover this false-positive case.

- **Bug: Grid column headers missing prefix** — `GridGroup` builds column headers using `getItemTextToDisplay()` only. Prefix and `_prefix` rendering extensions are not shown in grid column headers.

leoniedickson (approved):

PR #1900, opened 2026-04-16
Issue/1801

clinnygee:


leoniedickson (commented):
One thing to have a look at?

### `removeEmptyAnswers.ts` has the same bug

`removeEmptyAnswers.ts:146-151` has an identical item reconstruction pattern that also strips `definition`:

```ts
return {
  linkId: qItem.linkId,
  ...(qItem.text && { text: qItem.text }),
  ...(childQrItems.length > 0 && { item: childQrItems }),
  ...(updatedAnswers.length > 0 && { answer: updatedAnswers })
};
```

`getResponse()` itself doesn't call `removeEmptyAnswersFromResponse`, so it's not affected. But `removeEmptyAnswersFromResponse` is a publicly exported function and a realistic follow-on step — if a caller does `removeEmptyAnswersFromResponse(q, getResponse())`, definitions get stripped again. The same one-line fix should be applied there too.

PR #1922, opened 2026-04-27
Add clone response feature

### LINKED ISSUE #1880: Clone and create new response feature
On the "Existing Responses" page in the Smart Forms App you can either create response or select an existing response and open it. There is a need to be able to clone an existing response and use its content to populate a new response.
 
**Describe the solution you'd like**
Choosing to clone and create new response would copy the content of the existing response and populate a new QuestionnaireResponse. The elements like "id", "meta", and "authored" would not be copied as these are not populated intially for a new response. It would also mean prepopulation would not occur. The user would be required to use the "Re-populate form" function to update the form with patient record data. 

**Describe alternatives you've considered**
Current alternatives require re-entry of data.

**Additional context**
Existing response page where you can select an existing response to which a clone and create new button could be added.

Image

leoniedickson:
Adds a "Clone response" action to the dashboard responses table, the in-renderer existing responses table, and the viewer operations bar. Cloning strips id/meta/authored and resets status to in-progress so the clone is saved as a new response. Pre-population is skipped on load via router state.


clinnygee (approved):
LGTM

PR #1939, opened 2026-05-22
Issue/1926

### LINKED ISSUE #1926: Duplicate Condition onsetDateTime Causes Form Pre-population Failure
**Describe the bug**
The bug appears when multiple Condition resources are returned with the same: onsetDateTime

**To Reproduce**

The form population then fails with:

Form partially populated, there might be pre-population issues. View console for details.

Likely issue: the pre-population logic is not handling multiple matching Condition records with the same onset date correctly.

index-Df6BuxNt.js:616 
{
    "severity": "warning",
    "code": "invalid",
    "details": {
        "text": "toString: The input collection contains multiple items"
    }
}

**Expected behavior**
No error

leoniedickson:
## Summary

- `readPopulationExpressions` was adding children of `itemPopulationContext` repeat groups to the global `initialExpressions` map, causing `generateExpressionValues` to pre-evaluate them against the full resource collection. FHIRPath functions like `toString()` that require a single-item input would then throw, producing a spurious `OperationOutcomeIssue` warning. Population itself was never broken — the per-item evaluation in `constructRepeatGroupInstances` always worked correctly and never used the pre-computed value.
- Fix: exclude repeat group children from the global map in `readPopulationExpressions`, and read expressions directly from the questionnaire item in `constructRepeatGroupInstances` so they are always evaluated against the correctly scoped per-item context.
- Also adds explicit `@jest/globals` imports to all `sdc-populate` test files to resolve IDE type errors (matches the pattern already used in `smart-forms-renderer`).

### DIFF
--- a/packages/sdc-populate/src/SDCPopulateQuestionnaireOperation/test/addDisplayToCodings.test.ts
+++ b/packages/sdc-populate/src/SDCPopulateQuestionnaireOperation/test/addDisplayToCodings.test.ts
@@ -15,6 +15,7 @@
  * limitations under the License.
  */
 
+import { beforeEach, describe, expect, it, jest } from '@jest/globals';
 import { getCodeSystemLookupPromise } from '../api/lookupCodeSystem';
 import { resolveLookupPromises } from '../utils/resolveLookupPromises';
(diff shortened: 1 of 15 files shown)


clinnygee (approved):
LGTM. Good job.

PR #1957, opened 2026-06-03
changes to allow max screen width on renderer

clinnygee:


leoniedickson (commented):
Good work on the responsive rendering changes — the approach of uncapping the container and constraining individual items is sound. A few things worth addressing before merge:

**1. `RepeatItem` refactor causes label misalignment with adjacent non-repeat items**

Moving `RemoveItemButton` outside `ItemFieldGrid` into a sibling `Stack` makes `ItemFieldGrid` narrower by the button width, so its label column no longer aligns with adjacent non-repeat items (e.g. Name/Relationship next to Phone/Email in the Carers and Key Contacts section of the GPCCMP form). Two options:
- **Right offset**: give the `Box` wrapping `ItemFieldGrid` a fixed right padding equal to the button width — simple but fragile if button size ever changes
- **Absolute positioning**: drop the `Stack`, keep the `Box` at `maxWidth: lg`, and position `RemoveItemButton` absolutely to the right outside the box — cleaner, `ItemFieldGrid` stays the full `lg` width

**2. Severe duplication — `maxWidth` should live in one place**

The same `sx={{ maxWidth: (theme) => theme.breakpoints.values.lg }}` is copy-pasted across 25+ components. The natural home is `FullWidthFormComponentBox` in `Box.styles.ts` — one change there handles every consumer without touching 25 files (e.g. if the cap ever changes from `lg` to `xl`, all 25+ files need updating).

**3. `RepeatField.tsx` is now dead code**

`RepeatItem.tsx` no longer imports it, but the file remains. `RepeatItemContainerStack` in `RepeatItem.styles.tsx` is also only used by `RepeatField`, so both can be deleted.

**4. `DisplayItem` missed**

`DisplayItem.tsx` uses `FullWidthFormComponentBox` but wasn't given a `maxWidth` constraint. Likely an oversight — worth confirming whether display items should be capped too.

leoniedickson:
The misalignment issue (first point):
Screenshot 2026-06-09 at 11 46 00 am

clinnygee:
Good points. The DisplayItem had been left, i spoke to Liam yesterday and he did say he wanted it with the same treatment.

leoniedickson:
One remaining issue with the `RepeatItem` fix — on small screens (narrower than 1200px), the inner `maxWidth: lg` box fills the full container width, so `right: 0` of the outer relative box lands on top of the field content.

Moving `position: relative` to the inner box and using a negative right offset fixes both cases:

```tsx
<Box sx={{ position: 'relative', maxWidth: (theme) => theme.breakpoints.values.lg }}>
  <ItemFieldGrid ... />
  <Box sx={{ position: 'absolute', right: -40, top: '50%', transform: 'translateY(-50%)' }}>
    <RemoveItemButton ... />
  </Box>
</Box>
```

- **Small screen**: inner box fills the container, button sits just outside its right edge — no overlap
- **Large screen**: inner box is capped at 1200px, button sits just outside that edge

The `-40` is a hardcoded magic number tied to the button size, but it's one value in one place.

leoniedickson:
@clinnygee Sorry to push back another time - I'm seeing the misaligned fields again

Screenshot 2026-06-18 at 3 05 37 pm

clinnygee:
hey @leoniedickson  I don't have these issues on mine - lets talk tomorrow and see what is up

clinnygee:
@leoniedickson turns out i was running a different branch locally oops.. anyways it should be fixed now

leoniedickson (approved):
lgtm!

Issue #1971, opened 2026-06-16
For `date` / `dateTime` items, the renderer evaluates value constraints but never surfaces them at the field


oliveregger:
## Summary

For `date` / `dateTime` items, the renderer evaluates value constraints but never surfaces them at the field:

1. **`targetConstraint`** — a constraint whose `location` resolves to a date item is evaluated (it shows up in the aggregate validation / `OperationOutcome`), but the `human` message is **not shown inline** next to the date field.
2. **`minValue` / `maxValue`** — these extensions are **not enforced at all** on date items (only `integer` / `decimal` are supported), even though `checkMinValue` / `checkMaxValue` already contain working `date` / `dateTime` comparison branches.

Both come down to the date item components using a date-specific validation hook that bypasses the shared validation path.

- Package: `@aehrc/smart-forms-renderer` 1.3.1
- Commit: `4564f7ff`

## Reproduction

A `date` item (`linkId: dateOfBirth`) that should be limited to `[1900-01-01, today()]`.

**a) Dynamic upper bound via `targetConstraint` (Questionnaire-level extension):**

```json
{
  "url": "http://hl7.org/fhir/StructureDefinition/targetConstraint",
  "extension": [
    { "url": "key", "valueId": "dateOfBirthRange" },
    { "url": "severity", "valueCode": "error" },
    { "url": "expression", "valueExpression": {
        "language": "text/fhirpath",
        "expression": "%resource.descendants().where(linkId='dateOfBirth').answer.value.where($this < @1900-01-01 or $this > today()).exists()"
    }},
    { "url": "human", "valueString": "Date of birth must be between 1900-01-01 and today." },
    { "url": "location", "valueString": "Questionnaire.descendants().where(linkId='dateOfBirth')" }
  ]
}
```

Enter `1899-12-31` or a future date. **Expected:** the `human` message appears under the date field (as it does for `string` / `integer` items). **Actual:** the constraint evaluates (`isInvalid = true`, visible in the validation summary), but nothing is rendered inline at the field.

> Note: the `location` resolution is fine — `Questionnaire.descendants().where(linkId='dateOfBirth')` returns the item (verified with the bundled fhirpath). The gap is purely in the date field component, see below.

**b) Static bounds via `minValue` / `maxValue`:**

```json
{ "url": "http://hl7.org/fhir/StructureDefinition/minValue", "valueDate": "1900-01-01" }
```

[...]

PR #1974, opened 2026-06-18
feat(#1927): block save-as-final when form has validation errors; fix targetConstraint

### LINKED ISSUE #1927: Mandatory Code Elements Not Enforced (Condition, MedicationStatement, AllergyIntolerance, Immunization)
**Describe the bug**
The form allows creation of clinical resources without required code elements, resulting in invalid FHIR resources that violate profile constraints.

**To Reproduce**
It is currently possible to create:

Condition without code (1..1)
MedicationStatement without medication[x] (1..1)
AllergyIntolerance without code (1..1)
Immunization without vaccineCode (1..1)

Type in a value and backspace. 

**Expected behavior**
The system should:

Enforce mandatory fields (1..1) before allowing create
Prevent submission of invalid resources
Provide validation feedback to the user

Actual Behaviour

Resources are created with missing mandatory elements
Leads to invalid FHIR payloads and downstream issues (e.g. population errors)

### LINKED ISSUE #1954: targetConstraint expression logic opposite of what is expected
**Describe the bug**
It appear as though [targetConstraint](https://hl7.org/fhir/extensions/5.3.0-ballot-tc1/StructureDefinition-targetConstraint.html) has been implemented to trigger displaying why the invariant rules have not been met when the expression evaluates as true. [Constraints should always evaluate as true](https://build.fhir.org/conformance-rules.html#:~:text=A%20FHIRPath%20expression%20that%20must%20evaluate%20to%20true%20when%20run%20on%20the%20element), otherwise the invariant rules will be enforced.

Smart Forms implementation is documented [here](https://smartforms.csiro.au/storybook/index.html?path=/story/sdc-10-1-4-behavior-other-control--target-constraint-simple).

**To Reproduce**
Copy the Questionnaire from the storybook documentation above and paste into the playground. Input values to trigger constraint rules.

**Expected behavior**
The expression should normally evaluate to true and enforce rules when not true.

leoniedickson:
## Summary

- Blocks save-as-final (and save-as-amendment) when the form has validation errors — navigates to the first tab containing an error, highlights required fields, and shows a snackbar
- Same gate added to the viewer's "Save as Final" button for in-progress responses
- Fixes `targetConstraint` — the feature was entirely non-functional due to four compounding bugs; fixing it is a prerequisite for the save gate relying on `invalidItems`
- Fixes validation feedback not appearing on `date`/`dateTime` fields
- Fixes validation feedback squishing horizontally in grid cells for autocomplete choice fields

### Save gate — renderer (`SaveAsFinalAction.tsx`)

When `responseHasErrors` is true, save-as-final (and save-as-amendment) now:
1. Calls `highlightRequiredItems()` to surface required-field errors
2. Walks the questionnaire item tree to find the first tab containing an invalid item and navigates there via `switchTab()`
3. Shows a snackbar: _"The form has errors that must be fixed before saving as final."_
4. Aborts — extraction is not triggered

Both the write-back and non-write-back code paths are gated. New utility `renderer/utils/tabNavigation.ts` — `findFirstErrorTabIndex()` — with 12 unit tests.

### Save gate — viewer (`ViewerSaveAsFinal.tsx`)

The viewer's "Save as Final" button (shown for in-progress responses) now checks `responseHasErrors` before opening the confirmation dialog. If errors are present it calls `highlightRequiredItems()` and shows a snackbar: _"This response has errors. Edit the response to view and fix them before saving as final."_

### `responseHasErrors` added to store

`questionnaireResponseStore` gains a `responseHasErrors: boolean` derived from whether any OperationOutcome issue has `severity: 'error' | 'fatal'`. This distinguishes hard errors from warnings so the save gate doesn't block when only warnings are present. Updated in all six state-setting locations.

`validateTargetConstraint()` now also passes `targetConstraint.severityCode` and `targetConstraint.human` through to the OperationOutcome, replacing the previous hardcoded `severity: 'error'`.

### `targetConstraint` bug fixes (fixes #1954)

Four bugs were making the feature entirely non-functional:

1. **Item-level constraints never extracted** — `extractTargetConstraints` only read `questionnaire.extension`; most constraints live on `item.extension`. Added recursive traversal of `questionnaire.item`.

2. **`targetConstraintLinkIds` map not populated for item-level constraints** — `readTargetConstraintLocationLinkIds` only mapped constraints with a `location` extension. Added an `else if (targetConstraint.linkId)` branch for the common case.

3. **Expression logic inverted** — `isInvalid = result[0]` should be `isInvalid = !result[0]`. Per the FHIR spec, constraint expressions return `true` when the constraint is **met** (valid). The inversion meant spec-compliant expressions never showed errors. Also fixed the `intersect()` edge case where fhirpath.js returns `[]` for `false`.

   > **Breaking change:** questionnaires using expressions written in the old (inverted) convention (returning `true` when violated) will stop working. Expressions must return `true` when valid.

4. **Validation filter silently dropped item-level constraints** — the synthetic `target-constraint-${key}` fallback key doesn't exist in `itemMap`, causing the visibility filter to drop every item-level constraint from `filteredInvalidItems`. Implicitly resolved by bug 1: storing `linkId` on the `TargetConstraint` object means the real linkId is now used.

### Date/dateTime validation feedback

`CustomDateItem` and `CustomDateTimeItem` called `useDateValidation` directly, which only validates date format and has no awareness of target constraints, required fields, or `invalidItems`. Both components now call `useValidationFeedback` first and fall back to format feedback, matching the pattern used by other field types.

This resolves point 1 of #1971 (targetConstraint errors not displaying on date fields). Point 2 of that issue (`minValue`/`maxValue` not enforced on date items) is not addressed here.

### Validation feedback squishing in grid cells (choice select fields)

Autocomplete-based choice components were rendering feedback as a flex sibling of the Autocomplete, squishing it horizontally in grid cells. Replaced `StyledRequiredTypography` with `FormControl` + `FormHelperText` in `ChoiceSelectAnswerOptionFields`, `ChoiceSelectAnswerValueSetFields`, `OpenChoiceSelectAnswerOptionField`, and `OpenChoiceSelectAnswerValueSetField`. Error border added via `error={!!feedback}` on `StandardTextField`.

## Test plan

- [ ] Open a questionnaire with required fields — leave some empty, click "Save as Final" — gate fires, first error tab is selected, required fields highlighted
- [ ] Fill all required fields — "Save as Final" proceeds normally
- [ ] Questionnaire with `targetConstraint` expressions (returns `true` when valid) — violate constraint, confirm error shown; fix it, confirm error clears; save-as-final blocked then allowed
- [ ] In-progress response in viewer with missing required fields — "Save as Final" shows snackbar, does not open dialog
- [ ] Date/dateTime field with targetConstraint — confirm error feedback appears
- [ ] Autocomplete choice field in a grid — confirm validation feedback stacks vertically below the input

### DIFF
--- a/apps/smart-forms-app/src/features/renderer/components/RendererActions/SaveAsFinalAction.tsx
+++ b/apps/smart-forms-app/src/features/renderer/components/RendererActions/SaveAsFinalAction.tsx
@@ -32,6 +32,10 @@ import RendererSaveAsFinalOnlyDialog from './RendererSaveAsFinalOnlyDialog.tsx';
 import RendererSaveAsFinalWriteBackDialog from './RendererSaveAsFinalWriteBackDialog.tsx';
 import { populateQuestionnaire } from '@aehrc/sdc-populate';
 import { fetchResourceCallback } from '../../../prepopulate/utils/callback.ts';
+import { useSnackbar } from 'notistack';
+import CloseSnackbar from '../../../../components/Snackbar/CloseSnackbar.tsx';
+import { formHasErrorsMessage } from '../../../../interfaces/snackbar.interface.ts';
+import { findFirstErrorTabIndex } from '../../utils/tabNavigation.ts';
 
 interface SaveAsFinalActionProps extends SpeedDialActionProps {
   isSpeedDial?: boolean;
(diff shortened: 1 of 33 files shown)


clinnygee (approved):
lgtm

Issue #1994, opened 2026-07-03
Clearing an autocomplete choice field leaves a stub answer in the QuestionnaireResponse


leoniedickson:
## Description

When a user clears an autocomplete choice field using the **×** button, the field appears empty but the previous answer is not actually removed from the `QuestionnaireResponse`. A stub answer `[{ id }]` — containing only the internal answer key, no value — remains in the response item. This causes two problems:

1. **Stale data in the response** — the item persists in the QR with no value attached, rather than being removed.
2. **Required validation never fires** — the stub answer is treated as a valid answer, so the field is never flagged as empty.

## Reproduction (playground)

1. Build a form using **Aboriginal and Torres Strait Islander Health Check** and prepopulate.
2. Navigate to the **Medical History and Current Problems** tab.
3. In the **New Problems/Diagnoses** section, add a condition (e.g. _Asthma_).
4. Open Advanced Props → QR → Updatable Response and search for "asthma" — it should be present.
5. Clear the condition using the **×** button on the autocomplete field.
6. Search again in the Updatable Response — _asthma is still present_ despite the field appearing empty.

The validation side is also visible: with the field required, open the **Invalid Items** tab in Advanced Props after clearing — the item does not appear there because the stub answer causes `qrItemHasItemsOrAnswer` to return `true`, suppressing the required error.

## Root cause

Two functions interact incorrectly:

**1. `createEmptyQrItem`** (`packages/smart-forms-renderer/src/utils/qrItem.ts:89`)

When `answerKey` is defined (always true for choice fields), it returns:
```js
{ linkId, text, answer: [{ id: answerKey }] }
```
The answer array has one entry containing only the key ID — no `valueCoding`, `valueString`, etc. Both `ChoiceAutocompleteItem.tsx` and `ChoiceSelectAnswerOptionItem.tsx` call `createEmptyQrItem(qItem, answerKey)` in their clear/null-value handlers, emitting this stub to the store.

**2. `updateQrItemsInGroup`** (`packages/smart-forms-renderer/src/utils/qrItem.ts:111`)

Has two paths that both preserve the stub instead of removing the item:
- **Add-at-end path** (line 139): calls `qrItemHasItemsOrAnswer(newQrItem)` — returns `true` because `answer.length > 0`, so the stub is pushed.
- **Replace-at-position path** (line 149): checks `newQrItem.answer?.length` directly — also truthy, so the stub replaces the existing item rather than splicing it out.

[...]

PR #1997, opened 2026-07-08
fix: clearing autocomplete choice fields leaves stale answers in Ques…

clinnygee:
…tionnaireResponse (issue #1994)

Three related fixes for cleared fields persisting in the updatableResponse:

- OpenChoiceAutocompleteItem: the clear button fires onChange with reason "clear", which matched no branch in handleValueChange - the clear was swallowed and onQrItemChange never called, so the old valueCoding stayed in the response (the primary cause in the Aboriginal Health Check, whose condition fields are open-choice). Emit an empty item on newValue === null.

- updateQrItemsInGroup / qrItemHasItemsOrAnswer: a stub answer [{ id }] emitted by createEmptyQrItem(qItem, answerKey) no longer counts as an answer. Cleared items are now removed (not replaced by the stub) and stubs are never added on the empty-group/add-at-end/insert paths. New answerHasValue() helper checks for an actual value[x] or nested items.

- RepeatItem: empty repeat answer instances are now tracked in local state (same pattern as RepeatGroup/GroupTable) instead of being stored as value-less stub answers in the response.

Required validation now also treats value-less stub answers as empty, so cleared required fields are correctly flagged as invalid.

Adds end-to-end jsdom regression tests covering choice and open-choice autocompletes across standalone, repeat group, repeats:true and gtable contexts.

### DIFF
--- a/packages/smart-forms-renderer/src/components/FormComponents/OpenChoiceItems/OpenChoiceAutocompleteItem.tsx
+++ b/packages/smart-forms-renderer/src/components/FormComponents/OpenChoiceItems/OpenChoiceAutocompleteItem.tsx
@@ -129,9 +129,12 @@ function OpenChoiceAutocompleteItem(props: BaseItemProps) {
       setInput(newValue);
     }
 
+    // newValue is null when the field is cleared e.g. via the clear button ("clear") or Escape ("clearOnEscape") -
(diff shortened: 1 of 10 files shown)


leoniedickson (approved):
lgtm!

PR #1998, opened 2026-07-08
fix(sdc-assemble): use recursively-assembled child when flattening nested modular questionnaires

oliveregger:
## Problem

`assembleQuestionnaire` (packages/sdc-assemble/src/utils/assemble.ts) recurses into each subquestionnaire but **discards the result**:

```ts
for (let subquestionnaire of subquestionnaires) {
  const assembled = await assembleQuestionnaire(subquestionnaire, …, /* isRoot */ false, …);
  if (assembled.resourceType === 'OperationOutcome') return assembled;
  subquestionnaire = assembled;   // ← reassigns the loop variable, NOT the array element
}
…
const items = getItems(subquestionnaires); // reads the ORIGINAL, un-assembled children
```

Because `subquestionnaire = assembled` only rebinds the loop variable (and `assembleQuestionnaire` works on a `cloneDeep`, so the original is untouched), `getItems(subquestionnaires)` reads the **un-assembled** children. A subquestionnaire that is itself modular (nested / recursive modular questionnaire) is therefore spliced into its parent with its own `subQuestionnaire` placeholders **left unresolved** — nested modular never fully flattens.

## Fix

Iterate by index and write the assembled child back into the array, with a guard for the possibly-`undefined` element under `noUncheckedIndexedAccess`:

```ts
for (let i = 0; i < subquestionnaires.length; i++) {
  const subquestionnaire = subquestionnaires[i];
  if (!subquestionnaire) continue;
  const assembled = await assembleQuestionnaire(subquestionnaire, …, false, …);
  if (assembled.resourceType === 'OperationOutcome') return assembled;
  subquestionnaires[i] = assembled;
}
```

## Test

The existing `should handle recursive assembly of nested subquestionnaires` test only asserted `resourceType` / `version` / `url` / call-count, so it passed despite the bug. It is strengthened to assert that the deep child's item actually resolves into the output (`deep-question` present, `deep-subq-ref` placeholder gone). That assertion **fails on `main`** and passes with this change.

- Full `sdc-assemble` package suite: **123/123 passing**.

### DIFF
--- a/packages/sdc-assemble/src/test/assemble.test.ts
+++ b/packages/sdc-assemble/src/test/assemble.test.ts
@@ -33,6 +33,17 @@ const mockFetchQuestionnaireCallbackConfig = {
   sourceServerUrl: 'https://example.com/fhir'
 };
 
+// Collect every linkId in an item tree (depth-first), used to assert that (nested)
+// subQuestionnaire placeholders were resolved into their children's items.
+function collectLinkIds(items: Questionnaire['item']): string[] {
+  const linkIds: string[] = [];
+  for (const item of items ?? []) {
+    linkIds.push(item.linkId);
+    linkIds.push(...collectLinkIds(item.item));
+  }
+  return linkIds;
+}
+
 describe('assemble', () => {
   const createBaseQuestionnaire = (): Questionnaire => ({
     resourceType: 'Questionnaire',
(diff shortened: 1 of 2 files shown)


leoniedickson (approved):
Thanks for picking this up @oliveregger, looks good!

PR #1999, opened 2026-07-10
fix: scope repeating group validation errors to the invalid instance (#1985)

### LINKED ISSUE #1985: Validation errors on repeating group items show on all instances when only one is invalid
## Problem

`invalidItems` in the questionnaire response store is keyed by `qItem.linkId`. In a repeating group, every instance of a child item shares the same `linkId`. If one instance is invalid, `invalidItems[linkId]` is set and every rendered instance of that item shows the error — including instances with valid answers.

**Example:** a repeating group with a required text field. If the third instance is left empty and "Save as Final" is attempted, all instances show "This field is required", not just the empty one.

**Affected validation types:**
- **Required** — if one instance has no answer, all instances show the required error
- **Format/range** (regex, min/max, etc.) — if one instance has an invalid answer, all instances show the format error
- **`targetConstraint`** — not covered by the fix below; see note

## Effect on save gate

The save gate introduced in #1974 is unaffected — `responseHasErrors` correctly reflects whether *any* instance has errors. This is a display-only bug: errors appear on instances that don't deserve them after `highlightRequiredItems()` is called.

## Proposed fix

Set a unique `id` on each child `QuestionnaireResponseItem` within repeat group instances — in `useInitialiseRepeatGroups` for existing QR data, and wherever new repeat instances are created.

Validation would then key `invalidItems` by `qrItem.id ?? qItem.linkId`, and `useValidationFeedback` (which receives `qrItem`) would look up the same key. Items outside repeating groups have no `id` set and fall back to `linkId` — no behaviour change for non-repeating items.

**Files in scope:** `validate.ts`, `useValidationFeedback`, repeat group initialisation hooks.

## Note on `targetConstraint` in repeating groups

`targetConstraint` expressions are evaluated once against the entire `QuestionnaireResponse` as a document, returning a single `true`/`false` for the whole QR. There is no per-instance result — the constraint has no way to indicate which specific repeat instance violated it. Re-keying `invalidItems` by `qrItem.id` doesn't help here because the single `isInvalid` flag would still be stored under one key with no instance granularity.

Fixing this would require scoping the expression evaluation to each repeat instance separately — evaluating the constraint once per instance against that instance's subtree of the QR. This warrants a follow-up issue.

clinnygee:
Fixes #1985

## Problem
Validation errors on repeating group items showed on **all** instances when only one was invalid. Errors were stored/looked up in a flat `Record<linkId, OperationOutcome>`, and every field looked itself up as `invalidItems[qItem.linkId]`. Because all instances of a repeat-group child share one `linkId`, they collided on a single key — one invalid instance lit up all of them. The save gate was unaffected; this was purely a display issue.

## Fix
Errors are now keyed by `linkId` **plus the path of enclosing repeat-group instance indices** (e.g. `child-field///1`). Non-repeating items have an empty path, so their key stays a bare `linkId` — no behaviour change.

Validation (a pure function over the QR) and the display hook build the **identical** key from one shared helper, so they always agree without mutating or persisting anything into the `QuestionnaireResponse`. The instance-index path reaches descendants through a small React context provided per repeat instance / gtable row.

- **New** `utils/validateErrorKey.ts` — `getValidationErrorKey()` / `getBaseLinkIdFromErrorKey()`
- **New** `contexts/RepeatGroupInstanceContext.ts` — carries the instance-index path
- `utils/validate.ts` — threads `repeatInstancePath` through the recursion and keys all three error write sites via the helper
- `RepeatGroupItem.tsx` and `GroupTableRow.tsx` — provide their index path via the context
- `useValidationFeedback.ts` — reads the context and looks up the instance-scoped key (no signature change, so the ~25 leaf components are untouched)
- `SaveAsFinalAction.tsx` — maps error keys back to base linkIds so "jump to first error tab" still resolves

### The index must be the QuestionnaireResponse index, not the rendered index

This is the subtle part, and the first pass of this PR got it wrong.

Validation numbers instances by their position in the **QuestionnaireResponse**. The components providing the context originally appended their **rendered** index. Those two diverge whenever an instance is absent from the QR:

- `getGroupTableItemsToUpdate` drops gtable rows that are **unselected or empty**, while those rows stay rendered.
- `RepeatGroup` drops instances whose `qrItem` is still `null`.

So with three gtable rows where only the third is missing a required answer, deselecting the first row shifted every later row's key by one — validation wrote `family-name///1` while the offending row looked up `family-name///2`. The error vanished from the invalid row and appeared on a valid one, which is worse than the original bug (that at least included the right instance).

Both providers now derive the QR index from the same projection they mirror:

- `getQrRepeatGroupInstanceIndex` (`utils/repeatGroup.ts`)
- `getQrGroupTableRowIndex` (`utils/groupTable.ts`)

Each returns `null` for an instance with no QR counterpart, and `appendRepeatInstanceIndex` turns that into `UNMATCHABLE_REPEAT_INSTANCE_INDEX` (`-1`) — a key validation can never write, so an untouched instance stays silent instead of claiming index 0's error.

`repeatGroupIndex` is unchanged and still drives removal, answer changes, `useReadOnly` and `parentRepeatGroupIndex`; only the validation key uses the new index.

For gtable the skew was directly reachable by a user. For standard repeat groups it is currently latent — the Add Item button is disabled while the last instance is empty, which keeps `null` instances at the end — but the derivation is correct either way.

### Why not the literal `qrItem.id ?? linkId` from the issue?
The primary reported case is an **empty required field**, which has no `qrItem` at the leaf (it arrives as `undefined`) — so there'd be nothing to carry an id. A position-path key handles that case, needs no QR mutation/stripping, and composes for nested repeat groups.

## Out of scope (follow-ups)
- `targetConstraint` per-instance validation (explicitly deferred in the issue). Note this is **not** a regression from the re-keying: `useValidationFeedback` reads the `targetConstraints` store directly, before the `invalidItems` lookup, so targetConstraint display is untouched. The bare-linkId entry it writes into `invalidItems` is only consumed by the save gate and the tab jump, both of which are key-agnostic.
- Repeating **single** (non-group) items — still keyed by `linkId`

Nested repeat groups and `gtable` rows **are** covered by this change.

## Testing
- `test/validateErrorKey.test.ts` — key construction, round-trip, and the unmatchable-index behaviour
- `test/validateRepeatInstance.test.ts` — end-to-end `validateForm`: only the empty instance is flagged, not the answered one nor the bare `linkId`
- `test/utils/repeatGroup.test.ts` and the new cases in `test/utils/groupTable.test.ts` — each index helper asserted against the projection it mirrors, so the two can't drift apart
- **`stories/testing/RepeatInstanceValidation.stories.tsx`** — rendered-layer coverage, which is the layer where the index skew was invisible to the pure-function tests. One story per shape; the gtable story fails against the previous rendered-index behaviour (error appeared on row 1 instead of row 2).

Renderer `npm run build`: clean · repo `npm run lint`: 0 errors · app `tsc --noEmit`: clean
Renderer Jest: **2191 passed** (94 suites) · Storybook test-runner: **212 passed** (44 suites)


leoniedickson:
Nice write-up on the rendered-vs-QR index skew — that's a sharp catch, and the `UNMATCHABLE_REPEAT_INSTANCE_INDEX` sentinel is a clean way to keep an orphaned row silent instead of stealing another row's error.

Two small perf follow-ups, neither of which requires touching the approach:

1. **O(n²) index lookup** — `getQrRepeatGroupInstanceIndex`/`getQrGroupTableRowIndex` re-scan and filter the preceding slice on every call, and they're called once per row inside `repeatGroups.map(...)` (`RepeatGroupView.tsx:114,165`, and the gtable equivalent). For repeat groups/gtables with large instance counts this makes rendering quadratic. Could precompute the QR-index mapping once per parent in a single linear pass and index into it per row instead.

2. **Context identity churn** — `repeatInstancePath` in `RepeatGroupItem.tsx:66` (and `GroupTableRow.tsx`) is rebuilt via spread on every render before going into `RepeatGroupInstanceContext.Provider`, so every descendant consuming it via `useValidationFeedback` re-renders even when the path hasn't changed. Wrapping `appendRepeatInstanceIndex(...)` in `useMemo` keyed on `[parentRepeatInstancePath, qrInstanceIndex]` would fix this.

One doc nit:

3. The `invalidItems` comment in `questionnaireResponseStore.ts:50` still says it's keyed by bare `linkId` — worth updating to mention the `linkId///path` format now that repeat-instance keys are scoped, so a future change doesn't accidentally reintroduce the collision this PR fixes.

PR #2000, opened 2026-07-10
fix(#1931): show user-friendly warning when answerOption terminology lookup fails

### LINKED ISSUE #1931: answerOption list being displayed is a list of codes when $lookup fails
**Describe the bug**
When a Questionnaire has an answerOption list of codings that do not include displays, the renderer will attempt to lookup the displays from the terminology server to list in the UI. If the terminology server is down, it will list the codes. Listing codes is not appropriate for the UI. Would be better to display an error. 

**To Reproduce**
I noticed this when the terminology server was down and I [launched the GP CCMP](https://smartforms.csiro.au/launch?iss=https%3A%2F%2Fproxy.smartforms.io%2Fv%2Fr4%2Ffhir&launch=WzAsImZvcmVtYW4tY2F0ZXJpbmEiLCJwcmltYXJ5LXBldGVyIiwiIiwwLDAsMCwibGF1bmNoIG9wZW5pZCBmaGlyVXNlciBvbmxpbmVfYWNjZXNzIHBhdGllbnQvQWxsZXJneUludG9sZXJhbmNlLmN1cyBwYXRpZW50L0NvbmRpdGlvbi5jdXMgcGF0aWVudC9FbmNvdW50ZXIuciBwYXRpZW50L0ltbXVuaXphdGlvbi5jcyBwYXRpZW50L01lZGljYXRpb24uciBwYXRpZW50L01lZGljYXRpb25TdGF0ZW1lbnQuY3VzIHBhdGllbnQvT2JzZXJ2YXRpb24uY3MgcGF0aWVudC9QYXRpZW50LnIgcGF0aWVudC9RdWVzdGlvbm5haXJlUmVzcG9uc2UuY3J1cyB1c2VyL1ByYWN0aXRpb25lci5yIGxhdW5jaC9xdWVzdGlvbm5haXJlP3JvbGU9aHR0cDovL25zLmVsZWN0cm9uaWNoZWFsdGgubmV0LmF1L3NtYXJ0L3JvbGUvbmV3IiwiaHR0cHM6Ly9zbWFydGZvcm1zLmNzaXJvLmF1LyIsImE1N2Q5MGUzLTVmNjktNGI5Mi1hYTJlLTI5OTIxODA4NjNjMSIsIiIsIiIsIiIsIiIsMCwxLCJbe1wicm9sZVwiOlwiaHR0cDovL25zLmVsZWN0cm9uaWNoZWFsdGgubmV0LmF1L3NtYXJ0L3JvbGUvbmV3XCIsXCJ0eXBlXCI6XCJRdWVzdGlvbm5haXJlXCIsXCJjYW5vbmljYWxcIjpcImh0dHA6Ly93d3cuaGVhbHRoLmdvdi5hdS9hc3Nlc3NtZW50cy9HUENocm9uaWNDb25kaXRpb25NYW5hZ2VtZW50UGxhblwifSx7XCJyZWZlcmVuY2VcIjpcIlByYWN0aXRpb25lclJvbGUvZ3AtcHJpbWFyeS1wZXRlclwifV0iLCJodHRwczovL3Byb3h5LnNtYXJ0Zm9ybXMuaW8vdi9yNC9maGlyIixmYWxzZSwiIl0) form. An example set is in the "Patient details" tab, "Carer and key contacts" section, "Type" question item.

**Expected behavior**
Should display an appropriate error.

MaryamMehd:
## Summary

Fixes #1931

When a questionnaire has `answerOption` codings without `display` values, the renderer calls `CodeSystem/$lookup` to fetch human-readable labels. If the terminology server is unavailable, the lookups silently fail and the UI displays raw SNOMED codes — confusing for end users.

**Changes:**

- `addDisplayToAnswerOptions` now returns a `lookupFailedCodingKeys: Set<string>` (keyed by `system|code`) alongside the resolved options, tracking failed lookups at the individual coding level.
- `answerOptionsLookupFailures` is threaded through `QuestionnaireModel`, `QuestionnaireStoreType`, and `useQuestionnaireStore` so field components can detect which specific options could not be resolved.
- `ChoiceSelectAnswerOptionFields` and `OpenChoiceSelectAnswerOptionField` now **filter out** unresolvable options from the dropdown (so no raw codes are shown) and display an amber helper text: *"Some items in this list were not able to be displayed"*.
- Unit tests in `addDisplayToCodings.test.ts` updated to assert `lookupFailedCodingKeys` is correctly populated.

## Before / After

| Scenario | Dropdown behaviour |
|---|---|
| Working terminology server | Human-readable labels as normal ✅ |
| Server unavailable — **before fix** | Raw SNOMED codes (`133932002`, `394738000`) shown in dropdown ❌ |
| Server unavailable — **after fix** | Unresolvable options hidden; amber warning: "Some items in this list were not able to be displayed" ✅ |

## Test plan

- [ ] Run unit tests: `cd packages/smart-forms-renderer && npx jest --testPathPattern="addDisplayToCodings"` — all tests pass.
- [ ] In the Playground, load a questionnaire with `answerOption` codings that have no `display`, set the terminology server to a URL that validates but lacks SNOMED (e.g. `https://hapi.fhir.org/baseR4`), build the form — affected options are hidden and the amber warning appears.
- [ ] With a working terminology server, the dropdown shows proper labels with no warning.


leoniedickson:
Picked this PR back up and made a few changes on top of the original fix. Summary of what changed:

**Implementation now matches the description.** The original fix disabled the dropdown / showed an error alert when a lookup failed. Per the PR description (options should be filtered rather than shown as raw codes), it now filters out any option whose display couldn't be resolved, instead of showing it as a bracketed code or blocking the whole field.

**Simplified the failure-tracking.** Removed the `answerOptionsLookupFailures: Set<string>` threaded through the store/model/hooks - a missing `display` after `addDisplayToAnswerOptions` runs *is* the failure signal, so tracking it separately was redundant. Replaced with a single `isDisplayUnavailable(option)` check.

**Extended the fix's scope.** The original PR only touched the Select (`Autocomplete`) variant of `Choice`. This now covers Radio and Checkbox as well, and both `Choice` and `OpenChoice` item types - six components in total.

**Fixed a deeper "answered-option desync" bug class found while extending the above.** Previously-answered options were either leaking raw codes (Select) or vanishing entirely (Radio/Checkbox) when their terminology display couldn't be freshly re-resolved on reload (e.g. answer a question while the terminology server is up, reload while it's down). Answers now keep their originally-recorded display via the stored `QuestionnaireResponseItemAnswer`, falling back to being dropped (never a raw code) if no display exists anywhere.

**Fixed several related correctness bugs** surfaced along the way:
- Falsy-value matching bugs where `valueInteger: 0` / `valueString: ''` answers were treated as unset in the matching/comparison helpers.
- A checkbox "display-drift" bug where a terminology rename (display changed, `system`+`code` unchanged) would silently uncheck an already-answered option, because the equality check compared the whole `Coding` (including `display`) instead of just its identity (`system`+`code`).
- A key-collision bug in `answerOptionsToggleExpressions` matching, introduced by an earlier fix in this same effort.
- Made `OpenChoice` Select's "prefer live vs. stored display" behaviour consistent with `Choice` Select (they diverged pre-existing this PR).

**Cleanup / reuse.** Consolidated the duplicated "unavailable display" warning JSX and visibility-derivation logic (repeated across all 6 components) into a shared `useAnswerOptionVisibility` hook and `AnswerOptionUnavailableWarning` component, unified coding-equality checks on one `generateCodingKey` function, and optimised the options/answers merge from O(n×m) to O(n+m).

**Added test coverage** for the lookup-failure filtering and the answered-option desync scenarios across all six component variants, plus regression tests for the falsy-value and key-collision fixes.

**Out of scope, documented for follow-up** (not part of this PR):
- A pre-existing rendering bug where `valueInteger: 0` / `valueString: ''` options never render at all in Radio/Checkbox (a `0`-based scale would be missing its `0` option) - same root cause as one of the matching bugs above, but one layer further out in the rendering gate.
- Raw-code display in a few `apps/smart-forms-app`-level surfaces (repopulate/write-back previews, and notably the persisted `QuestionnaireResponse.text.div` narrative generation) - app-level, not renderer-level, and needs its own scoping decision.

clinnygee (approved):
LGTM

PR #2061, opened 2026-07-28
Keep html-react-parser out of the core extension reader

### LINKED ISSUE #2058: Core extension reader pulls in html-react-parser for one call site
`src/utils/extensions.ts` imports `html-react-parser` (line 27) and calls `htmlParse` at exactly one site (line 343), inside `getTextDisplayFlyover`.

That puts a DOM parser, and a function that returns a React element, into the import graph of every consumer that reads item extensions. A consumer that only wants to inspect a Questionnaire's extensions gets the parser too. It also blocks marking the package `sideEffects: false` later, and it is the kind of edge that makes the engine hard to reuse outside a browser.

The split is available at a natural data boundary: the lookup itself is pure traversal (walk `childItem`s, check the `display` type and the `flyover` item control, prefer a `rendering-xhtml` extension over `item.text`), and only the final step needs a parser.

### Suggested direction

Have `utils/extensions.ts` return the raw xhtml string plus the plain text, and let the UI layer parse it. `src/hooks/useParseXhtml.tsx` already owns xhtml parsing for other item text, so it is the natural home for the parsing wrapper.

### Scope

This is one edge, not the whole story. `src/utils/choice.ts:27` still carries a runtime import into the React layer (`generateCodingKey` and `generateOptionKey` from `hooks/useAnswerOptionsToggleExpressions`), so `src/utils` is not DOM-free after this change alone. Worth addressing separately if there is appetite; this issue is only about the extension reader and its parser.

KyleOps:
Any code that reads a Questionnaire's extensions currently also pulls in an HTML parser, even when it never renders anything. This moves `getTextDisplayFlyover`, the only function in `utils/extensions.ts` that parses XHTML, into `hooks/useParseXhtml.tsx`, which already owns XHTML parsing. `utils/extensions.ts` becomes a pure deletion: the function and the `html-react-parser` import are removed, nothing else changes. It also removes one of the two things blocking the package from being marked tree-shakeable later.

Fixes #2058

The function moved wholesale rather than being split into a data half and a parsing wrapper: it has a single consumer (`hooks/useRenderingExtensions.ts`), and the traversal only needs `isSpecificItemControl` and `getXHtmlString`, which remain exported from `utils/extensions.ts`. A wholesale move keeps the parser out of the core extension reader with the smallest reviewable diff.

For review: the function body in `useParseXhtml.tsx` is identical to what was deleted from `extensions.ts` (only the doc comment changed), and the five pre-existing flyover tests moved verbatim from `extension.test.tsx` to `useParseXhtml.test.tsx`, plus one new test pinning the fall-through to a later flyover childItem.

One consumer-visible note: `getTextDisplayFlyover` was never exported from the package barrel, but deep imports from `lib/utils/extensions` would break; a compatibility re-export would reintroduce the removed dependency.

Testing: full renderer jest suite green (92 suites), eslint and prettier clean on touched files. `src/utils/choice.ts` still has one other hooks-layer import; left for a follow-up. No dependency or lockfile changes, so no conflicts with the sibling PRs.


clinnygee:

This PR merged into `main` cleanly, no conflicts.

Results on the combined branch:

| Check | Result |
| --- | --- |
| Renderer `tsc` build | clean |

Manual testing against that branch also looked good.

### One thing worth making explicit

The PR description presents the relationship to #2063 as a stacked base branch. It is stronger than that — **#2061 is a hard functional dependency of #2063, and has to land first.**

#2063's `engineEntrypoint.test.ts` asserts that `html-react-parser` never enters the headless engine's runtime import graph, and its `engine.ts` re-exports `isSpecificItemControl` / `getDecimalPrecision` from `utils/extensions`. Until this PR moves `getTextDisplayFlyover` out of that module, the parser is still in the closure and that test fails.

So this is not just base-branch bookkeeping — merging #2063 without this one would land a red test. Worth noting for whoever sequences the merges.

PR #2062, opened 2026-07-28
Make the terminology transport injectable

### LINKED ISSUE #2060: Terminology transport is hardcoded to fhirclient with no seam for a consumer
Three functions in the renderer each construct their own `fhirclient` inline to reach a terminology server:

- `utils/valueSet.ts` `getValueSetPromise()` for `ValueSet/$expand`
- `utils/valueSet.ts` `validateCodePromise()` for `ValueSet/$validate-code`
- `utils/questionnaireStoreUtils/addDisplayToCodings.ts` `getCodeSystemLookupPromise()` for `CodeSystem/$lookup`

There is no seam for a consumer to supply its own transport. A host that already has an authenticated terminology client, its own retry and logging policy, or a non-standard auth flow cannot route the engine's terminology traffic through it. The practical result is that such a host bypasses the engine's value set handling altogether and reimplements the parts it needs, which means `answerValueSet` bindings, `$validate-code` and `addDisplayToCodings` stop being useful to it.

`resolveValueSetPromises()` also carries a hardcoded `const timeoutMs = 5000` with no way to change it, which is a poor fit for a slow or metered connection.

### There is already a pattern for this in the repo

`@aehrc/sdc-populate` and `@aehrc/sdc-assemble` both take injected fetch callbacks rather than owning a transport. So this is consistency with the sibling packages rather than a new idea, and the callback shape `sdc-populate` already defines (`FetchTerminologyCallback`) could be reused directly, letting one implementation serve both.

### Suggested direction

An optional transport callback plus a configurable timeout, defaulting to exactly today's behaviour so existing consumers notice nothing. `terminologyServerStore` already owns the terminology server URL, so it is the natural place for it.

KyleOps:
Fixes #2060

Three functions each built their own terminology HTTP client inline (`ValueSet/$expand` and `ValueSet/$validate-code` in `valueSet.ts`, `CodeSystem/$lookup` in `addDisplayToCodings.ts`), with a 5 second timeout hardcoded at one of them. All three now go through a single seam, the new `terminologyRequest()` in `src/utils/terminologyRequest.ts`, so an adopter with their own authentication, retry, or caching policy can use the engine's terminology features instead of working around them.

The injection API: `terminologyServerStore.setRequestOptions({ fetchTerminologyCallback, requestTimeoutMs })`, in the store where the terminology server URL already lives. The callback type is `@aehrc/sdc-populate`'s existing `FetchTerminologyCallback`, re-exported from the renderer, so one implementation serves both packages. With nothing injected, the seam uses the same `fhirclient` transport and query strings as before; default behaviour is unchanged. `requestTimeoutMs` (default 5000) covers only the build-time batch `$expand` of `answerValueSet` bindings, the one place a timeout exists today.

The second commit, which wires `ItemRepopulateButton` to prefer the injected callback when handing a transport to `sdc-populate`, is separable: it is the only behaviour change to a shipped component in this PR and it has no test coverage. It is kept as its own commit so it can be dropped without affecting the seam.

Testing: 16 new unit tests covering the default path (exact pre-change query strings for all three operations), the injected path, error propagation, and the timeout. Full renderer suite, `tsc --noEmit`, and the build are green. Not exercised against a real terminology server.

Follow-up candidate, not part of this PR: the renderer now has two default transports (the `fhirclient` seam and the fetch-based `fetchTerminologyCallback` in `api/callback.ts`); consolidating on one is worth its own issue.

### DIFF
--- a/documentation/docs/dev/storehooks/terminology-server-store/index.mdx
+++ b/documentation/docs/dev/storehooks/terminology-server-store/index.mdx
@@ -7,6 +7,8 @@ sidebar_position: 5
 The TerminologyServerStore exposes the following properties:
 
 - **url**
+- **fetchTerminologyCallback**
+- **requestTimeoutMs**
 
 Here's how you use the store hook in your React functional component:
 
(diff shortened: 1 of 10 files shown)


clinnygee:

Results on the combined branch:

| Check | Result |
| --- | --- |
| Renderer `tsc` build | clean |

Manual testing against that branch also looked good, including the per-item repopulate path.

I also checked the seam is actually complete: after this PR there is exactly **one** `fhirclient` import left in the whole renderer (`utils/terminologyRequest.ts` itself), and the autocomplete path reaches it too — `useTerminologyServerQuery` goes through `getValueSetPromise`. So no terminology call site is left bypassing the injection point.

### Needs a rebase — two trivial conflicts with `main`

GitHub currently shows this as conflicting. Both are adjacency conflicts with the i18n work that landed after the branch was cut (#1995), and both resolve by keeping **both** sides:

- `src/index.ts` — the new `FetchTerminologyCallback` re-export sits next to the new `rendererStrings` exports
- `ItemRepopulateButton.tsx` — `injectedFetchTerminologyCallback` sits next to the new `rendererStrings` selector

No semantic overlap, nothing to reconcile.

### Suggestion (lands in #2063, noting it here for context)

Once `/engine` from #2063 is in the same tree, a headless consumer can call `terminologyServerStore.setRequestOptions({...})` but can't **name** its argument — `TerminologyRequestOptions` and `FetchTerminologyCallback` are exported from the root barrel by this PR, but `engine.ts` only re-exports `terminologyServerStore` and `TerminologyServerStoreType`.

That's a bit of a shame, because this seam is arguably the single most important API for a `/engine` consumer: the built-in `fhirclient` transport needs a browser `fetch`, so injection is what makes the headless entrypoint genuinely usable on React Native. Detail is on #2063; nothing to change here.

### Minor observation on the follow-up you flagged

Related to the "two default transports" follow-up in the description: the app's own populate path (`apps/smart-forms-app/src/features/prepopulate/utils/callback.ts`) defines its own `fetchTerminologyCallback` and passes it directly to `populateQuestionnaire()`, so it never consults the store and is unaffected by this seam.

Not a problem for this PR — default behaviour is unchanged either way — but it does mean an adopter who injects via `setRequestOptions()` still won't capture the app-level populate traffic. Might be worth folding into the same consolidation issue.

### One type-level breaking change worth calling out in the changelog

`TerminologyServerStoreType` gains four required members (`fetchTerminologyCallback`, `requestTimeoutMs`, `setRequestOptions`, `resetRequestOptions`). Purely additive for anyone *consuming* the store, but it breaks anyone who *implements* the interface — most likely a hand-rolled test mock typed as `TerminologyServerStoreType`. Niche, but a silent `tsc` break on upgrade if unmentioned.

PR #2063, opened 2026-07-28
Add a headless engine entrypoint and an exports map

### LINKED ISSUE #2059: No supported headless entrypoint, and no exports map
`@aehrc/smart-forms-renderer` publishes `main: lib/index.js` with no `exports` map and no `sideEffects` field.

Two consequences:

**There is no supported way to use the engine without the UI.** Importing anything from the barrel pulls in `@mui/material`, `@emotion/*`, `react-dom`, `react-markdown` and both drag-and-drop libraries. A consumer that only needs the form engine (a React Native host, a test, a service) has to deep-import internal paths instead, for example:

```ts
import { buildForm, getResponse } from '@aehrc/smart-forms-renderer/lib/utils/manageForm';
import { questionnaireStore } from '@aehrc/smart-forms-renderer/lib/stores/questionnaireStore';
import { canBeObservationExtracted } from '@aehrc/smart-forms-renderer/lib/utils/extractObservation';
```

Those paths are not public API and can move in a patch release, so a consumer relying on them has to pin an exact version and carry its own tests to detect the UI layer creeping back into its bundle.

**Bundlers cannot tree-shake the package.** With no `exports` map and no `sideEffects` flag, a web app that uses part of the renderer still pays for the rest.

### Suggested direction

An `exports` map with a curated headless entrypoint (something like `./engine`) exporting the vanilla Zustand stores and the DOM-free utils, alongside the unchanged barrel at `.`.

One caveat worth stating up front, because it is easy to get wrong: an `exports` map is a **whitelist**, so adding one to a package that never had one silently breaks every existing deep import unless a passthrough is included. A single `./lib/*` pattern is not sufficient either, because an `exports` pattern has no directory-index fallback, so paths that resolve today via a directory `index.js` would stop resolving.

`sideEffects: false` would need #2053 first, since the barrel currently has a webfont side effect.

KyleOps:
Fixes #2059

Server-side and React Native consumers need the form engine (stores, `buildForm`, populate and extract plumbing) without the DOM-bound UI tree. This adds `@aehrc/smart-forms-renderer/engine`, a headless entrypoint whose runtime import graph contains no UI package, and the package's first `exports` map. The two ship together because the clean `/engine` specifier only exists through the map.

The engine barrel is a strict subset of the root barrel: nothing becomes public by appearing in `/engine` alone, and a test enforces the subset. `engineEntrypoint.test.ts` cruises the graph with dependency-cruiser: no UI library may load at runtime, and the type surface a consumer must resolve to typecheck is pinned to three `@mui/material` entries, needed because `RendererConfig` keys breakpoints by MUI's augmentable `Breakpoint` type. Getting the pin that small meant moving the `RenderingExtensions` interface into `interfaces/` and importing `UseResponsiveProps` from its own module rather than the hooks barrel, so type edges no longer drag the hooks layer (and with it `html-react-parser`) into the closure.

A second guard, `packageExports.test.ts`, keeps the exports map in sync with the source tree: every directory barrel under `src/` must have a `./lib/<dir>` entry, since the map has no directory-index fallback.

For review: `src/engine.ts` (the surface), the `exports` map in `package.json`, and the two test files. Stacked on #2061 (base branch is `issue/2058`).

Testing: full renderer jest suite green (94 suites, 2201 tests). Every published `./lib` subpath was resolution-checked under node16, bundler, and node10 settings when the map was authored; the guard test keeps it in sync from here.

Not verified: Metro (React Native) resolution and an `npm pack` install into a real consumer.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18874,120 +18874,6 @@
         "node": ">= 0.8"
       }
     },
-    "node_modules/dependency-cruiser": {
-      "version": "18.2.0",
-      "resolved": "https://registry.npmjs.org/dependency-cruiser/-/dependency-cruiser-18.2.0.tgz",
-      "integrity": "sha512-xMDoLD0no6pDInR8/4rIIqZ4mERDnsjezk8PkNORYSfBLvjCOogUxaruepmi1uQtZQlYUgdT2u7G3jTlgKqNjw==",
-      "dev": true,
-      "license": "MIT",
-      "dependencies": {
-        "acorn": "8.18.0",
-        "acorn-jsx": "5.3.2",
-        "acorn-jsx-walk": "2.0.0",
-        "acorn-loose": "8.5.2",
-        "acorn-walk": "8.3.5",
-        "commander": "15.0.0",
-        "enhanced-resolve": "5.24.5",
-        "ignore": "7.0.6",
-        "interpret": "3.1.1",
-        "is-installed-globally": "1.0.0",
-        "json5": "2.2.3",
-        "picomatch": "4.0.5",
-        "prompts": "2.4.2",
-        "rechoir": "0.8.0",
-        "safe-regex": "2.1.1",
-        "semver": "7.8.5",
-        "tsconfig-paths-webpack-plugin": "4.2.0",
-        "watskeburt": "6.0.0"
-      },
-      "bin": {
-        "depcruise": "bin/dependency-cruise.mjs",
-        "depcruise-baseline": "bin/depcruise-baseline.mjs",
-        "depcruise-fmt": "bin/depcruise-fmt.mjs",
-        "depcruise-wrap-stream-in-html": "bin/wrap-stream-in-html.mjs",
-        "dependency-cruise": "bin/dependency-cruise.mjs",
-        "dependency-cruiser": "bin/dependency-cruise.mjs"
-      },
-      "engines": {
-        "node": "^22||^24||>=26"
-      }
-    },
-    "node_modules/dependency-cruiser/node_modules/acorn": {
-      "version": "8.18.0",
-      "resolved": "https://registry.npmjs.org/acorn/-/acorn-8.18.0.tgz",
-      "integrity": "sha512-lGq+9yr1/GuAWaVYIHRjvvySG5/4VfKIvC8EWxStPdcDh/Ka7FG3twP6v4d5BkravUilhIAsG4Qj83t02LWUPQ==",
-      "dev": true,
-      "license": "MIT",
-      "bin": {
-        "acorn": "bin/acorn"
-      },
-      "engines": {
-        "node": ">=0.4.0"
-      }
-    },
-    "node_modules/dependency-cruiser/node_modules/commander": {
-      "version": "15.0.0",
-      "resolved": "https://registry.npmjs.org/commander/-/commander-15.0.0.tgz",
-      "integrity": "sha512-z67u4ZhzCL/Tydu1lJARtEZYWbWaN7oYLHbsuzocr6y4N6WZAagG3RQ4FW61V1/0+jImpj293XfrcYnd1qxtPg==",
-      "dev": true,
-      "license": "MIT",
-      "engines": {
-        "node": ">=22.12.0"
-      }
-    },
-    "node_modules/dependency-cruiser/node_modules/ignore": {
-      "version": "7.0.6",
-      "resolved": "https://registry.npmjs.org/ignore/-/ignore-7.0.6.tgz",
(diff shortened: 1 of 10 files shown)


clinnygee:

Only conflict merging this PR was `package-lock.json`, resolved by taking the branch's version and re-running `npm install`.

Results on the combined branch (after the `./lib/i18n` fix below):

| Check | Result |
| --- | --- |
| Renderer `tsc` build | clean, emits `lib/engine.js` + `lib/engine.d.ts` |

I also resolution-checked ~20 published subpaths against the new `exports` map with Node's resolver (root, `/engine`, `package.json`, every enumerated `./lib/<dir>` barrel, nested paths like `lib/utils/questionnaireStoreUtils/addDisplayToCodings`, `.js`-suffixed requests, and `./src/*`). All resolve. The `./lib/*.js` vs `./lib/*` pattern ordering behaves correctly.

Manual testing against that branch also looked good.

---

### Blocker: needs a rebase, and the map is missing `./lib/i18n`

`src/i18n/index.ts` landed on `main` after this branch was cut (#1995, commit 8356f836). The `exports` map here doesn't have a `./lib/i18n` entry, and since the map has no directory-index fallback, `@aehrc/smart-forms-renderer/lib/i18n` silently stops resolving for consumers.

This PR's own guard catches it. Rebased onto current `main`, it's the only failure in the suite:

```
FAIL src/test/packageExports.test.ts
  ● package.json exports map › enumerates every published directory barrel under lib/

    - Array []
    + Array [ "i18n" ]
```

The guard doing exactly its job is a good sign for the design. The fix is four lines:

```json
"./lib/i18n": {
  "types": "./lib/i18n/index.d.ts",
  "default": "./lib/i18n/index.js"
},
```

With that added, the full suite goes green. I only have it on the local test branch — it needs to land here.

### Suggestion: `engine.ts` can't name the terminology injection API

This one only becomes visible once #2062 is in the same tree.

#2062 adds `terminologyServerStore.setRequestOptions({ fetchTerminologyCallback, requestTimeoutMs })` — the seam that makes the renderer's terminology traffic routable through a consumer's own HTTP client. That is arguably *the* API a headless `/engine` consumer needs most, since the built-in `fhirclient` transport wants a browser `fetch`.

`engine.ts` currently exports:

```ts
export type { TerminologyServerStoreType } from './stores/terminologyServerStore';
export { terminologyServerStore, useTerminologyServerStore } from './stores/terminologyServerStore';
```

So a `/engine` consumer can *call* `setRequestOptions({...})` but can't name its argument type or type their callback — `TerminologyRequestOptions` and `FetchTerminologyCallback` aren't re-exported from `/engine`.

Both are already in the root barrel (`src/index.ts` lines 45 and 53 after #2062), so adding them here keeps the subset guard satisfied:

```ts
export type { TerminologyRequestOptions } from './stores/terminologyServerStore';
export type { FetchTerminologyCallback, FetchTerminologyRequestConfig } from '@aehrc/sdc-populate';
```

Ordering-dependent on #2062, so happy for it to be a follow-up if you'd rather not couple the two.

### Note on sequencing

Flagged on #2061 too: that PR is a **hard functional dependency** of this one, not just a stacked base branch. `engineEntrypoint.test.ts` asserts `html-react-parser` stays out of the engine runtime graph, and `engine.ts` re-exports from `utils/extensions` — so without #2061 removing `getTextDisplayFlyover` from that module, this PR's own test fails.

KyleOps:
Rebased onto main now that #2061 has landed, and added the `./lib/i18n` entry. Your read was exactly right: the guard caught it, nothing else was missing, and it really was the four lines.

Where the rebased branch stands: renderer suite 96 suites / 2231 tests, root lint 0 errors, `apps/smart-forms-app` tsc plus vite build clean. I also re-ran the resolution check across the whole published surface rather than a sample, enumerating every specifier the built package exposes: 704 of 704 resolve, including `lib/i18n` and the eight internal paths our own consumer imports.

Heads up that I force-pushed, so your combined test branch will need re-pulling.

On the terminology re-exports, I want them but can't add them here. `TerminologyRequestOptions` does not exist until #2062 lands, so the export would not compile against main. Taking your offer of a follow-up and I will raise it as soon as #2062 merges. Agreed it is the API a headless consumer reaches for first, so it should not sit unexported for long.

I also added a short entry points section to the renderer README. A new public entrypoint with nothing documented leaves anyone reading the merged commit no way to find it. It records that `/engine` is a subset of the root barrel, that both entries are bundler targets rather than Node loadable, and that deep `lib/...` imports are a compatibility surface rather than API.

One gap still open, and worth someone's eyes before it is relied on: Metro. React Native resolves through Metro, not Node, and Metro has its own `exports` implementation with a history of gaps in subpath patterns. Everything I have checked used Node's resolver. It does not block this PR, since the map is additive under either resolver, but I would want it closed before anyone leans on `/engine` from a React Native host.

Your sequencing note was better than mine, incidentally. I had described #2061 as a stacked base; you are right that it is a hard functional dependency, since the engine graph test fails without it.

Anything you would word differently in the README section, or is it fine as is?

clinnygee (approved):
lgtm

PR #2065, opened 2026-07-31
Back the HAPI FHIR servers with RDS PostgreSQL

KyleOps:
## Problem

Both HAPI FHIR servers store everything in RAM and lose it on every task replacement.

Neither `forms-server` nor `ehr-proxy` configured a datasource, so HAPI fell back to its default in-memory database. Confirmed from the running container logs:

```
jdbc:h2:mem:test_mem
```

The task definitions had `"volumes": []` and `"mountPoints": []`, and there was no RDS instance or EFS filesystem in the account backing either server. The data existed only in the Fargate task's memory.

This is the cause of the reported "the server gets wiped and someone has to repopulate it by hand". It is not intermittent corruption, it is total loss on any task replacement. ECS does this routinely. From the `ehr-proxy` service events:

```
2026-07-30T15:30:59  (task 8ced774a...) (port 8080) is unhealthy ... (reason Request timed out)
2026-07-30T15:26:53  Amazon ECS replaced 1 tasks due to an unhealthy status
```

A second symptom of the same root cause: both services track `hapiproject/hapi:latest`, and because they restarted at different times they had silently drifted to different HAPI versions (`ehr-proxy` on 8.4.0, `forms-server` on 8.10.0).

## Change

- Each HAPI endpoint gets a dedicated RDS PostgreSQL instance in the private subnets of its existing VPC.
- HAPI is pointed at it via `SPRING_DATASOURCE_URL` / `_DRIVERCLASSNAME` / `_USERNAME` / `_PASSWORD` and the `HapiFhirPostgresDialect`.
- Credentials are generated into Secrets Manager and injected as ECS task **secrets**, so they never appear in the task definition.
- The database security group accepts connections only from the service that owns it.
- The image is pinned to `hapiproject/hapi:v8.10.0-3`.
- A five minute health check grace period is added.

No change was needed to either app stack: the construct takes the VPC from `cluster.vpc`, so the diff is contained to the two `hapi-endpoint` constructs plus docs.

### Sizing

Each database gets the memory its in-memory predecessor was borrowing from the task, keeping the two in the same proportion to their tasks:

| | Task | Database | Storage |
|---|---|---|---|
| `ehr-proxy` | 1 vCPU / 2 GiB | `db.t4g.micro` (1 GiB) | 20 GiB GP3, autoscale to 100 |
| `forms-server` | 2 vCPU / 4 GiB | `db.t4g.small` (2 GiB) | 20 GiB GP3, autoscale to 100 |

Burstable Graviton suits the intermittent demo and evaluation traffic these serve. Both are single-AZ, which matches the stated lack of uptime requirements. Encrypted, 7 days of automated backups, deletion protection on, and `RemovalPolicy.SNAPSHOT`.

### Versions

- **PostgreSQL 18.3**, the current RDS default in `ap-southeast-2` and the major with the longest support runway. Minor upgrades are automatic.
- **HAPI `v8.10.0-3`**, the newest published tag. Verified byte-identical to `latest` by digest, so pinning changes nothing today but stops unattended migrations against real data later.

## Verification

CDK synth of both stacks succeeds and produces the expected resources. Beyond that, the exact container and env var configuration in this PR was run locally against PostgreSQL 18.3:

- HAPI 8.10.0-3 boots and completes schema migration on a cold database (~30s).
- `PUT Questionnaire` and `POST QuestionnaireResponse` both return 201.
- Indexed search (`QuestionnaireResponse?questionnaire=...`) returns the expected match.
- **The container was then destroyed and recreated, and all data was still present.** This is the case that fails today.
- Second boot logs zero schema warnings.

The initially chosen `HapiFhirPostgres94Dialect` was caught during this testing as deprecated in HAPI 8.10 and replaced with `HapiFhirPostgresDialect`.

## Deployment notes

- **The servers will come up empty.** The current contents live only in RAM and cannot be migrated. `forms-server` currently holds 141 Questionnaires and `ehr-proxy` holds 15 Questionnaires / 11 QuestionnaireResponses / 51 Patients / 20 Practitioners, all of which will need reloading after this deploys.
- First deploy takes roughly 10 to 15 minutes for the RDS instances to become available.
- Deletion protection is on, so a future `cdk destroy` will need it disabled manually. That is deliberate.

## Follow-up, not in this PR

- Seeding and reset automation, so repopulating is a command rather than a manual task. `aehrc/smart-forms-fhir-data` already holds starter bundles and `aehrc/sparked-test-data-loader` already solves the load/clear problem.
- The two `hapi-endpoint` constructs are near-identical and this PR adds more duplication between them. Factoring them into one shared construct is worth doing, but deliberately kept out of a change whose purpose is fixing data loss.

### DIFF
--- a/deployment/ehr-proxy/README.md
+++ b/deployment/ehr-proxy/README.md
@@ -13,6 +13,12 @@ From: https://github.com/hapifhir/hapi-fhir-jpaserver-starter
 
 Base URL: `https://proxy.smartforms.io/fhir/`
 
+The endpoint is backed by a dedicated RDS PostgreSQL instance (`db.t4g.micro`) created alongside it
+in the same construct. HAPI defaults to an in-memory H2 database when no datasource is configured,
+which means the server starts empty every time the Fargate task is replaced, so the datasource is
+configured explicitly. Credentials are generated into Secrets Manager and injected into the task as
+`SPRING_DATASOURCE_USERNAME` and `SPRING_DATASOURCE_PASSWORD`.
+
 ### smart-proxy
 A microservice running in Docker that acts as a SMART on FHIR plugin layered on top of the HAPI server.
 It requires the `FHIR_SERVER_R4` environment variable pointing to the HAPI server.
--- a/deployment/ehr-proxy/hapi-endpoint/lib/index.ts
+++ b/deployment/ehr-proxy/hapi-endpoint/lib/index.ts
@@ -1,14 +1,24 @@
 // import * as cdk from 'aws-cdk-lib';
+import { Duration, RemovalPolicy } from 'aws-cdk-lib';
 import { Construct } from 'constructs';
+import { InstanceClass, InstanceSize, InstanceType, SubnetType } from 'aws-cdk-lib/aws-ec2';
 import {
   AwsLogDriver,
   Cluster,
   Compatibility,
   ContainerImage,
   FargateService,
+  Secret as EcsSecret,
   TaskDefinition
 } from 'aws-cdk-lib/aws-ecs';
 import { RetentionDays } from 'aws-cdk-lib/aws-logs';
+import {
+  Credentials,
+  DatabaseInstance,
+  DatabaseInstanceEngine,
+  PostgresEngineVersion,
+  StorageType
+} from 'aws-cdk-lib/aws-rds';
 
 // import * as sqs from 'aws-cdk-lib/aws-sqs';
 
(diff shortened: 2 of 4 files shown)


KyleOps:
PR to create and switch over to a extendal DB (RDS) so that data can persist (if you want it to) in between months and you can query it etc...

For actual loading and resetting of the test data for specific events Sparked has scripts that we could employ like https://github.com/aehrc/sparked-fhir-server-configuration/blob/825ef197822f321ceb0c0a87322464ced14919e9/scripts/clear_test_data.py#L330 + https://github.com/aehrc/sparked-fhir-server-configuration/blob/825ef197822f321ceb0c0a87322464ced14919e9/scripts/load_test_data.py#L96 OR we could use an automated app like https://loader.sparked-fhir.com/ 

CC: @liambarnes @edeati

leoniedickson (approved):

Issue #1777, opened 2025-11-05
Template-based extraction array bug (name.given)


jschoen-gefyra:
**Describe the bug**
I have this example Questionnaire:

```
{
  "resourceType": "Questionnaire",
  "id": "DemoTemplatebasedExtractionQuestionnaire",
  "meta": {
    "profile": [
      "https://gematik.de/fhir/isik/StructureDefinition/ISiKFormularDefinition"
    ]
  },
  "subjectType": [
    "Patient"
  ],
  "contained": [
    {
      "resourceType": "Patient",
      "id": "IsikPatientTemplate",
      "meta": {
        "profile": [
          "https://gematik.de/fhir/isik/StructureDefinition/ISiKPatient"
        ]
      },
      "name": [
        {
          "use": "official",
          "extension": [
            {
              "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractContext",
              "valueString": "item.where(linkId = 'name')"
            }
          ],
          "_text": {
            "extension": [
              {
                "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractValue",
                "valueString": "item.where(linkId = 'given' or linkId = 'family').answer.value.join(' ')"
              }
            ]
          },
          "_family": {
            "extension": [
              {
                "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractValue",
                "valueString": "item.where(linkId = 'family').answer.value.first()"
              }
            ]
          },
          "_given": [
            {
              "extension": [
                {
                  "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractValue",
                  "valueString": "item.where(linkId = 'given').answer.value"
                }
              ]
            }
          ]
        }
      ],
      "gender": "unknown",
      "_gender": {
        "extension": [
          {
            "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractValue",
            "valueString": "item.where(linkId = 'gender').answer.value.first().code"
          }
        ]
      },
      "_birthDate": {
        "extension": [
          {
            "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractValue",
            "valueString": "item.where(linkId = 'birthDate').answer.value.first()"
          }
        ]
      }
    }
  ],
[...]

PR #2093, opened 2026-08-26
fix(sdc-template-extract): merge templateExtractValue into the array element its path names

### LINKED ISSUE #2092: sdc-template-extract: templateExtractValue appends a new array element instead of populating the one its path names
**Describe the bug**

In `@aehrc/sdc-template-extract` (1.0.15), a `templateExtractValue` whose value path names an element that **already exists** in the template's array does not populate that element — the evaluated value is appended as a separate, half-populated element.

Root cause is in `walkTemplateAndInsertValue` (`packages/sdc-template-extract/src/SDCExtractQuestionnaireResponseOperation/utils/templateInsert.ts`):

```ts
current[finalSegment] = deepmerge(existingNode, valueToInsert);
```

`deepmerge`'s default array strategy is **concatenation**. `buildValuesToInsert` turns a value path like `Immunization.protocolApplied[0].doseNumberPositiveInt` into `{protocolApplied: [{doseNumberPositiveInt: 2}]}`, which is then concatenated onto the static `protocolApplied: [{targetDisease: …}]` rather than merged into index 0.

This appears to be distinct from the array work in 1.0.15 (#1777), which addressed insert positioning and index bookkeeping; the object-merge path was not touched.

**To Reproduce**

Only reproduces under a `templateExtractContext` — a gated array element is deleted and re-seeded from the static template, then deep-merged. Without a gate the value is assigned directly and the bug does not appear.

1. Template = a Bundle whose `entry[0]` carries a `templateExtractContext`.
2. `entry[0].resource` = an `Immunization` with `protocolApplied[0]` declared statically:
   ```json
   "protocolApplied": [{ "targetDisease": [{ "coding": [{ "code": "67924001" }] }] }]
   ```
3. Add a `templateExtractValue` on `protocolApplied[0]._doseNumberPositiveInt` reading an integer answer.
4. Run `$extract` with that answer set to `2`.

**Expected behavior**

```json
"protocolApplied": [{ "targetDisease": [...], "doseNumberPositiveInt": 2 }]
```

**Actual**

```json
"protocolApplied": [{ "targetDisease": [...] }, { "doseNumberPositiveInt": 2 }]
```

Besides being the wrong shape, this breaks cardinality for anything 1..1 within the element — here `Immunization.protocolApplied.doseNumber[x]`, which ends up absent from the first entry and orphaned in the second.

**Additional context**

Found while building an SDC template-based `$extract` for a Swiss public-health IG (CH EKM), where a vaccination row extracts `protocolApplied.targetDisease` statically and the dose count via a value path into the same element.

Happy to open a PR — I have a fix (an index-wise `arrayMerge`) plus a minimal test fixture, with the full suite passing.

oliveregger:
Fixes #2092

## Problem

A `templateExtractValue` whose value path names an element that **already exists** in the template's array does not populate that element. The evaluated value is appended as a separate, half-populated element instead.

Given a template with `protocolApplied[0]` declared statically:

```json
"protocolApplied": [{ "targetDisease": [{ "coding": [{ "code": "67924001" }] }] }]
```

and a value expression writing `protocolApplied[0].doseNumberPositiveInt`:

```
expected  "protocolApplied": [{ "targetDisease": [...], "doseNumberPositiveInt": 2 }]
actual    "protocolApplied": [{ "targetDisease": [...] }, { "doseNumberPositiveInt": 2 }]
```

The value path names index 0, so the value belongs in that element. Splitting it also breaks cardinality for anything that is 1..1 within the element — here `Immunization.protocolApplied.doseNumber[x]`, which ends up absent from the first entry and orphaned in the second.

## Cause

`walkTemplateAndInsertValue` deep-merges an evaluated object into the existing node when both are objects. `deepmerge`'s **default array strategy is concatenation**, so an array nested inside that object is appended to rather than merged.

`buildValuesToInsert` turns `protocolApplied[0].doseNumberPositiveInt` into `{protocolApplied: [{doseNumberPositiveInt: 2}]}`, which is then concatenated onto the static `protocolApplied: [{targetDisease: …}]`.

This only surfaces under a `templateExtractContext`: a gated array element is deleted and re-seeded from the static template, then deep-merged. Without a gate the value is assigned directly and the bug does not appear — which is why the test fixture uses a context-gated Bundle entry.

Note this is a different code path from #1777 / the array fixes in 1.0.15, which addressed insert positioning and index bookkeeping. The object-merge path was untouched.

## Fix

Pass an `arrayMerge` that merges by index, appending only elements beyond the target's length so genuinely new entries behave as before.

The explicit array branches above this one — `existingNode.push(valueToInsert)` and the array-to-array concat, which handle repeating values such as multiple `given` names — are not affected; `deepmerge` is only reached when both sides are non-array objects.

## Tests

- New fixture triple `QArrayElementMerge` / `QRArrayElementMerge` / `extractedArrayElementMerge`, plus one `describe` in `extract.test.ts`. Minimal, no existing goldens touched.
- First commit adds the failing test only (intentionally red, if CI runs per-commit); second commit applies the fix.
- Full suite after the fix: **70/70 passing**, `tsc --noEmit` clean, eslint clean.

Happy to squash the two commits if you would rather not have a red commit in history, and to swap the `any` types in `mergeArrayByIndex` for `deepmerge`'s own `ArrayMergeOptions` — they currently match the surrounding file's style.

## Notes

Found while building an SDC template-based `$extract` for a Swiss public-health IG (CH EKM), where a vaccination row extracts `protocolApplied.targetDisease` statically and the dose count via a value path into the same element.

### DIFF
--- a/packages/sdc-template-extract/CHANGELOG.md
+++ b/packages/sdc-template-extract/CHANGELOG.md
@@ -5,6 +5,10 @@ This log documents changes for [@aehrc/sdc-template-extract](https://www.npmjs.c
 
 Changelog only includes changes from version 1.0.6 onwards.
 
+## sdc-template-extract [Unreleased]
+#### Fixed
+- A `templateExtractValue` whose value path names an element that already exists in the template's array (e.g. `Immunization.protocolApplied[0].doseNumberPositiveInt`) now populates that element instead of appending a separate, half-populated one. Arrays nested inside a merged object are merged by index rather than concatenated.
+
 ## sdc-template-extract [1.0.15] - 2026-01-13
 #### Fixed
 - Fixed issues relating to the extraction of arrays, including issue [#1777](https://github.com/aehrc/smart-forms/issues/1777).
--- a/packages/sdc-template-extract/src/SDCExtractQuestionnaireResponseOperation/tests/extract.test.ts
+++ b/packages/sdc-template-extract/src/SDCExtractQuestionnaireResponseOperation/tests/extract.test.ts
@@ -3,6 +3,7 @@ import type { OutputParameters, ReturnParameter } from '../interfaces';
 import { createInputParameters } from '../utils/createInputParameters';
 import type {
   Bundle,
+  Immunization,
   Observation,
   Parameters,
   QuestionnaireResponseItem,
@@ -20,6 +21,7 @@ import { extractedRegularMedicationsModified } from './resources/extracted/extra
 import { extractedRegularMedicationsWithPatchAdd } from './resources/extracted/extractedRegularMedicationsWithPatchAdd';
 import { extractedComplexTemplateExtract } from './resources/extracted/extractedComplexTemplateExtract';
 import { extractedSepsisRisk } from './resources/extracted/extractedSepsisRisk';
+import { extractedArrayElementMergeImmunization } from './resources/extracted/extractedArrayElementMerge';
 
 // QuestionnaireResponses
 import { QRAllergiesAdverseReactions } from './resources/questionnaireResponses/QRAllergiesAdverseReactions';
@@ -32,6 +34,7 @@ import { QRRegularMedicationsModified } from './resources/questionnaireResponses
 import { QRRegularMedicationsWithPatchAdd } from './resources/questionnaireResponses/QRRegularMedicationsWithPatchAdd';
 import { QRComplexTemplateExtract } from './resources/questionnaireResponses/QRComplexTemplateExtract';
 import { QRSepsisRisk } from './resources/questionnaireResponses/QRSepsisRisk';
+import { QRArrayElementMerge } from './resources/questionnaireResponses/QRArrayElementMerge';
 
 // Questionnaires
 import { QAllergiesAdverseReactions } from './resources/questionnaires/QAllergiesAdverseReactions';
@@ -44,6 +47,7 @@ import { QRegularMedicationsModified } from './resources/questionnaires/QRegular
 import { QRegularMedicationsWithPatchAdd } from './resources/questionnaires/QRegularMedicationsWithPatchAdd';
 import { QComplexTemplateExtract } from './resources/questionnaires/QComplexTemplateExtract';
 import { QSepsisRisk } from './resources/questionnaires/QSepsisRisk';
+import { QArrayElementMerge } from './resources/questionnaires/QArrayElementMerge';
 import { parametersIsFhirPatch } from '../utils/typePredicates';
 
 // Mock the fetchQuestionnaire callback function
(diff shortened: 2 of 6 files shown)


leoniedickson:
Thanks for this — the root-cause writeup and fixture are clear, and the positional-merge direction is the right fix for the reported issue. While testing it I found a regression in the new `mergeArrayByIndex`, plus a fix for it.

## Duplicate repeating value gets silently dropped

`mergeArrayByIndex`'s fallback branch:

```ts
} else if (target.indexOf(item) === -1) {
  destination.push(item);
}
```

drops an item if it's already present *anywhere* in `target`, instead of only appending genuinely new ones. So a repeating value that happens to answer the same twice (e.g. a `given` name entered twice) silently loses the duplicate.

Repro — Questionnaire (static `prefix` alongside two other sibling values, to also exercise the fix below):

```json
{
  "id": "Combined",
  "url": "https://smartforms.csiro.au/docs/tests/Combined",
  "status": "draft",
  "subjectType": ["Patient"],
  "contained": [
    {
      "resourceType": "Patient",
      "id": "patTemplate",
      "name": [
        {
          "extension": [
            { "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractContext", "valueString": "item.where(linkId = 'name')" }
          ],
          "prefix": ["Dr"],
          "_family": {
            "extension": [
              { "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractValue", "valueString": "item.where(linkId = 'family').answer.value.first()" }
            ]
          },
          "_use": {
            "extension": [
              { "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractValue", "valueString": "item.where(linkId = 'use').answer.value.first()" }
            ]
          },
          "_given": [
            {
              "extension": [
                { "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractValue", "valueString": "item.where(linkId = 'given').answer.value" }
              ]
            }
          ]
        }
      ]
    }
  ],
  "item": [
    {
      "extension": [
        { "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtract",
          "extension": [{ "url": "template", "valueReference": { "reference": "#patTemplate" } }] }
      ],
      "linkId": "patient", "text": "Patient Information", "type": "group",
      "item": [
        { "linkId": "name", "text": "Name", "type": "group", "repeats": true,
          "item": [
            { "linkId": "family", "text": "Family/Surname", "type": "string" },
            { "linkId": "use", "text": "Name use", "type": "string" },
            { "linkId": "given", "text": "Given Name(s)", "type": "string", "repeats": true }
          ]
        }
      ]
    }
  ]
}
```

QuestionnaireResponse:

```json
{
  "resourceType": "QuestionnaireResponse",
  "status": "completed",
  "questionnaire": "https://smartforms.csiro.au/docs/tests/Combined",
  "item": [
    { "linkId": "patient", "item": [
      { "linkId": "name", "item": [
        { "linkId": "family", "answer": [{ "valueString": "Citizen" }] },
        { "linkId": "use", "answer": [{ "valueString": "official" }] },
        { "linkId": "given", "answer": [{ "valueString": "Alex" }, { "valueString": "Alex" }] }
      ]}
    ]}
  ]
}
```

Actual (current PR): `"prefix": ["Dr"], "given": ["Alex"]` — `given` drops the second entry.
Expected: `"given": ["Alex", "Alex"]`.

## Suggested fix

The `indexOf` guard exists because `combineStaticTemplateData` re-spreads the *entire* static template object into every sibling `templateExtractValue`'s payload within a context, not just once — so a static array field (`prefix` above) would otherwise duplicate itself once per sibling value. Gating the static-data lookup on `isNewInsert` removes that redundant re-spread at its source (a merge into an already-seeded element doesn't need it — `deepmerge` already preserves untouched keys), which means the guard has nothing left to defend against and can just be dropped:

```ts
// insertValuesToPath
const staticTemplateData = isNewInsert
  ? getStaticTemplateDataAtPath(entryPath, cleanTemplate, populateIntoTemplateWarnings)
  : {};
```
```ts
// mergeArrayByIndex — drop the guard
} else {
  destination.push(item);
}
```

Both parts are needed together — I checked: removing only the guard fixes `given` but makes `prefix` duplicate once per sibling value (`["Dr","Dr","Dr","Dr"]` on the repro above). With both changes, the repro gives `"prefix": ["Dr"], "given": ["Alex", "Alex"]` — both correct — and the full existing suite still passes (70/70, no regressions).

Happy to leave this to you to fold in, or push a commit if that's easier — your call.

## Unrelated, pre-existing bug (out of scope here)

While testing I also hit a separate bug: a repeating value gets split into separate array elements (in reverse order) when an earlier sibling `templateExtractValue` in the same context evaluates empty. Confirmed it's untouched by this PR and reproduces on `main` too — out of scope for this PR, just flagging it. Filed separately: aehrc/smart-forms#2107.

leoniedickson (approved):
Thanks for this fix! Merging as submitted — a couple of follow-up bugs in the new mergeArrayByIndex (covered in the review comment above) will land in a separate quick follow-up PR right after this.

PR #2096, opened 2026-08-28
fix (sdc-populate) `$populate` silently drops open-choice answers that are not in answerOption

### LINKED ISSUE #2095: `$populate` silently drops open-choice answers that are not in `answerOption`
# `$populate` silently drops open-choice answers that are not in `answerOption`

**Describe the bug**

During pre-population, `@aehrc/sdc-populate` drops `valueCoding` answers produced by an `initialExpression` when they do not match one of the item's `answerOption` codings - including on `open-choice` items, where FHIR allows answers outside the listed options.

 What makes this hard to deal with is that nothing surfaces. The expression resolves correctly and the answer is present in the intermediate response, then it disappears before the QuestionnaireResponse is returned, no error, no warning, no entry in the OperationOutcome. An answer that was produced and then discarded looks exactly like one that was never produced, so there is nothing to triage on and each occurrence has to be traced by hand.

**To Reproduce**

An `open-choice` item that lists common diagnoses as options but has to accept any ICD-10 code recorded on the patient:

```json
{
  "linkId": "dx-primary",
  "type": "open-choice",
  "answerOption": [
    {
      "valueCoding": {
        "system": "http://hl7.org/fhir/sid/icd-10-cm",
        "code": "E11.9",
        "display": "Type 2 diabetes mellitus without complications"
      }
    }
  ],
  "extension": [
    {
      "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-initialExpression",
      "valueExpression": {
        "language": "text/fhirpath",
        "expression": "%condition.code.coding.first()"
      }
    }
  ]
}
```

1. Populate it for a patient whose `Condition.code.coding.first()` is any ICD-10 coding other than `E11.9`, for example `J44.9`.
2. `dx-primary` comes back with no answer.

**Expected behavior**

`dx-primary` is answered with the populated coding. `open-choice` exists so that answers outside the listed options are allowed, and the same function already preserves non-matching `valueString` answers for exactly that reason - the coding path just misses the same consideration.

Plain `choice` items should keep filtering as they do today.

**Additional context**

There appear to be two separable defects behind this:

1. `filterAndNormaliseAnswers` (`utils/processValueSets.ts`) filters regardless of item type. `valueCoding` answers absent from the item's options - `answerOption`, expanded valueSet, or contained valueSet are dropped, and the item type is not available at that point, so `open-choice` cannot be distinguished from `choice`. The neighbouring `valueString` branch already carries the comment "to support open-choice questions where arbitrary strings are allowed".

2. `findInAnswerOptions` (`utils/answerOption.ts`) never matches Coding values. `parseValueToAnswer` passes raw FHIRPath results to it, but the signature is `(options, str: string)` and it compares with `===` against option codes and displays. Coding objects, anything from `%condition.code.coding.first()` and similar, never match, so a Coding that *is* among the options is not normalised to that option's coding. It survives today only because the later options filter happens to compare by `code`.

- Reproduced on `@aehrc/sdc-populate` 4.7.1.
- A fix for the first defect is behaviour-changing only for `open-choice` items; `choice` items keep their current filtering, so existing forms are unaffected.
- Found in a Da Vinci DTR prior-authorization form (CMS-0057 tooling) on an ICD-10 primary-diagnosis field: the `Condition` was passed correctly, the FHIRPath resolved correctly, and the answer vanished between the two.
- I have a working implementation for both defects with regression tests, and would be happy to submit a PR if this approach is acceptable.

not-in-stock:
## Summary

Fixes #2095.

`$populate` was silently dropping or rewriting pre-populated coding answers. The pattern behind every defect here is the same: population matches codings in two places (`findInAnswerOptions` during initial population, the valueSet answers filter in `processValueSets.ts`) with different rules, and each disagreement between the two produces a silently wrong answer.

One commit per defect, details in the commit messages; the reproduction is in #2095:

- open-choice answers outside answerOption/valueSet options were filtered out
- Coding values from FHIRPath results never matched answerOptions (string-only comparison)
- the valueSet answers filter matched by bare code: a same-code coding from a different system was rewritten to the option's coding, and display-only options matched by `undefined === undefined`; both steps now share one matcher
- a systemless Coding fell through to `valueString` with the object as its payload, a structurally invalid answer
- answerOption `{valueInteger: 0}` never matched (falsy guard)
- a value-less Quantity could pass as a Coding; `valueIsCoding` is now a property allowlist
- cleanup: open-choice items are recorded during the existing population walk instead of a second traversal, and the filter's new parameter is required instead of defaulting to empty

## A related renderer gap

The `valueCoding` outliers this PR preserves are visible in the select/autocomplete open-choice variants but not in the radio/checkbox variants: `OpenChoiceRadioAnswerOptionItem` and `useOpenLabel` surface out-of-option answers only via `valueString`, so a preserved coding renders as an unanswered item while the response still carries it. The gap predates this PR, but this change makes it reachable through `$populate` alone. I can file it as a separate issue if you agree it is renderer scope rather than population scope.

## Backwards Compatibility

- `choice` items keep the current filtering behaviour; only `open-choice` items keep out-of-option codings
- `findInAnswerOptions()` still accepts plain strings with unchanged matching; Coding support is additive
- System agreement applies only when both codings specify a system, so systemless data keeps matching as before
- The one intentional break is compile-time only: `filterValueSetAnswersRecursive()` gained a required parameter. All in-repo callers are updated; external callers get a type error, not a behaviour change

## Test Plan

- [x] Full `sdc-populate` suite passes, `tsc --noEmit` and prettier clean
- [x] Every new regression test was run against the unfixed code and observed to fail
- [x] End-to-end `populate()` test: an open-choice and a choice item prefilled from the same out-of-options Condition coding; the open-choice answer survives, the choice answer is filtered

### DIFF
--- a/packages/sdc-populate/src/SDCPopulateQuestionnaireOperation/test/answerOption.test.ts
+++ b/packages/sdc-populate/src/SDCPopulateQuestionnaireOperation/test/answerOption.test.ts
@@ -83,4 +83,37 @@ describe('findInAnswerOptions', () => {
     const result = findInAnswerOptions([], 'anything');
     expect(result).toBeUndefined();
   });
+
+  it('finds option by Coding value matching system and code', () => {
+    const result = findInAnswerOptions(options, { system: 'http://loinc.org', code: '1234-5' });
+    expect(result).toEqual({
+      valueCoding: { ...codingOption.valueCoding }
+    });
+  });
+
+  it('finds option by Coding value without a system when codes match', () => {
+    const result = findInAnswerOptions(options, { code: '1234-5' });
+    expect(result).toEqual({
+      valueCoding: { ...codingOption.valueCoding }
+    });
+  });
+
+  it('returns undefined for Coding value with a different system', () => {
+    const result = findInAnswerOptions(options, {
+      system: 'http://snomed.info/sct',
+      code: '1234-5'
+    });
+    expect(result).toBeUndefined();
+  });
+
+  it('returns undefined for Coding value with a non-matching code', () => {
+    const result = findInAnswerOptions(options, { system: 'http://loinc.org', code: '9999-9' });
+    expect(result).toBeUndefined();
+  });
+
+  it('does not match a Quantity-like object against coding options', () => {
+    const quantity = { value: 10, unit: '1234-5', system: 'http://loinc.org', code: '1234-5' };
+    const result = findInAnswerOptions(options, quantity as never);
+    expect(result).toBeUndefined();
+  });
 });
--- a/packages/sdc-populate/src/SDCPopulateQuestionnaireOperation/test/answerOption.test.ts
+++ b/packages/sdc-populate/src/SDCPopulateQuestionnaireOperation/test/answerOption.test.ts
@@ -117,3 +117,12 @@ describe('findInAnswerOptions', () => {
     expect(result).toBeUndefined();
   });
 });
+
+describe('findInAnswerOptions zero-valued integer options', () => {
+  it('matches answerOption valueInteger 0 for a populated string "0"', () => {
+    const options = [{ valueInteger: 0 }, { valueInteger: 5 }];
+
+    expect(findInAnswerOptions(options, '0')).toEqual({ valueInteger: 0 });
+    expect(findInAnswerOptions(options, '5')).toEqual({ valueInteger: 5 });
+  });
+});
--- a/packages/sdc-populate/src/SDCPopulateQuestionnaireOperation/test/answerOption.test.ts
+++ b/packages/sdc-populate/src/SDCPopulateQuestionnaireOperation/test/answerOption.test.ts
@@ -33,7 +33,7 @@
  */
 
 import { describe, expect, it } from '@jest/globals';
-import { findInAnswerOptions } from '../utils/answerOption';
+import { findInAnswerOptions, valueIsCoding } from '../utils/answerOption';
 
 describe('findInAnswerOptions', () => {
   const codingOption = {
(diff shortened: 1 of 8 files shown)


leoniedickson:
Thanks for tackling this — the unification via `codingMatchesOption` is a nice fix for the two-places-disagree root cause.

One related gap I found: `valueIsCoding` (`answerOption.ts:94`) requires `code` to be a `string`, so a Coding that legally omits `code` (e.g. `{system, display}` only — every `Coding` element is `0..1` in base FHIR) is never recognized as a Coding by `findInAnswerOptions` or `parseValueToAnswer` (`parse.ts:121`). It falls through to the generic-object fallback and gets flattened to `valueString`, losing `system` entirely, instead of `valueCoding` — even though `codingMatchesOption`'s own display-only branch (lines 113-116) treats a code-less Coding as legitimate and matchable.

**Repro**: a `choice` item with a display-only `answerOption`:
```json
{
  "answerOption": [
    { "valueCoding": { "system": "http://example.com/local-codes", "display": "Query positive" } }
  ]
}
```
populated from an `initialExpression` resolving to `{system: "http://example.com/local-codes", display: "Query positive"}` (no `code`) produces `{"valueString":"Query positive"}` instead of matching the option as `valueCoding`.

**Suggested fix** — allow `code` to be absent, not just a string:
```ts
export function valueIsCoding(value: unknown): value is Coding {
  return (
    typeof value === 'object' &&
    value !== null &&
    ((value as Coding).code === undefined || typeof (value as Coding).code === 'string') &&
    Object.keys(value).every((key) => CODING_PROPERTIES.has(key))
  );
}
```

Happy for you to fold this into the PR if it's easier from your end, or I can push a commit — whichever's more convenient.

not-in-stock:
Thanks @leoniedickson, good catch. The display-only branch in `codingMatchesOption` was indeed unreachable for real FHIRPath results because `valueIsCoding` rejected the value before it got there.

I'll fold the fix into this PR.
One adjustment on top of your suggestion: with `code` optional, objects like `{}` or `{userSelected: true}` also start passing the property allowlist and would end up as an empty `valueCoding` in the response, so the predicate will additionally require at least one of `system` or `code`. A bare `{display}` object keeps its current fallback. Will add a regression test for your repro along with it.

leoniedickson (approved):
Thanks @not-in-stock, looks good

PR #2109, opened 2026-09-07
fix(renderer): add a clear button to slider items (#2094)

### LINKED ISSUE #2094: Slider item does not include a Clear button by default
**Describe the bug**
For one of our Assessment forms, we are using a slider for Pupil Size.
We noticed that the slider by default does not include a Clear button.

**To Reproduce**
The sample is below, it consist of 2 questions:
1. Pupil Size (mm)
2. Reaction

For 'Reaction', If you select one of the radio buttons (e.g. No Reaction), you will get a 'CLEAR' button which can be used to clear the answer
However for 'Pupil Size (mm)', if I move the slider across to say 4, there is no 'CLEAR' button.
User is confused and thinks using the 'CLEAR' button on 'Reaction' should also clear the 'Pupil Size' slider.

       {
          "item": [
            {
              "code": [
                {
                  "code": "leftEyePupilDiameter",
                  "system": "https://fhir.medirecords.com/CodeSystem/mr-parameter-name",
                  "display": "Left eye Pupil diameter Manual"
                },
                {
                  "code": "1",
                  "system": "https://fhir.medirecords.com/CodeSystem/mr-parameter-type",
                  "display": "Physical"
                }
              ],
              "text": "Pupil Size (mm)",
              "type": "integer",
              "linkId": "awsLeftPupilSize",
              "extension": [
                {
                  "url": "http://hl7.org/fhir/StructureDefinition/questionnaire-itemControl",
                  "valueCodeableConcept": {
                    "text": "A control where an axis is displayed between the high and low values and the control can be visually manipulated to select a value anywhere on the axis.",
                    "coding": [
                      {
                        "code": "slider",
                        "system": "http://hl7.org/fhir/questionnaire-item-control",
                        "display": "Slider"
                      }
                    ]
                  }
                },
                {
                  "url": "http://hl7.org/fhir/StructureDefinition/questionnaire-sliderStepValue",
                  "valueInteger": 1
                },
                {
                  "url": "http://hl7.org/fhir/StructureDefinition/minValue",
                  "valueInteger": 0
                },
                {
                  "url": "http://hl7.org/fhir/StructureDefinition/maxValue",
                  "valueInteger": 10
                },
                {
                  "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-observationExtract",
                  "valueBoolean": true
                }
              ]
            },
            {
              "code": [
                {
                  "code": "leftPupilPupillaryResponse",
                  "system": "https://fhir.medirecords.com/CodeSystem/mr-parameter-name",
                  "display": "Left pupil Pupillary response"
                },
                {
                  "code": "1",
                  "system": "https://fhir.medirecords.com/CodeSystem/mr-parameter-type",
                  "display": "Physical"
                }
              ],
              "text": "Reaction",
              "type": "choice",
              "linkId": "awsLeftPupilReaction",
              "repeats": false,
              "required": false,
              "extension": [
                {
                  "url": "http://hl7.org/fhir/StructureDefinition/questionnaire-itemControl",
                  "valueCodeableConcept": {
                    "coding": [
                      {
                        "code": "radio-button",
                        "system": "http://hl7.org/fhir/questionnaire-item-control"
                      }
                    ]
                  }
                },
                {
                  "url": "http://hl7.org/fhir/StructureDefinition/questionnaire-choiceOrientation",
                  "valueCode": "vertical"
                },
                {
                  "url": "http://hl7.org/fhir/StructureDefinition/minValue",
                  "valueInteger": 0
                },
                {
                  "url": "http://hl7.org/fhir/StructureDefinition/maxValue",
                  "valueInteger": 10
                },
                {
                  "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-observationExtract",
                  "valueBoolean": true
                }
              ],
              "answerValueSet": "#EyeReaction"
            }
          ],
          "text": "Left Pupil",
          "type": "group",
          "linkId": "awsLeftPupil-grp"
        }

**Expected behavior**
Can we have a 'CLEAR' button for sliders as well.
When pressed, it will clear the value of the slider, as if the question has not been answered.
It should not set it to value '0' because '0' is different to not set.

clinnygee:
Fixes #2094

## Problem

Slider items had no way to unset an answer. Once the user moved the thumb, the answer was there for good.

The reporter's assessment form puts a slider (`Pupil Size (mm)`) next to a radio group (`Reaction`). The radio group gets a `CLEAR` button; the slider does not. Users assumed the one `CLEAR` button covered both questions and were confused when the pupil size reading stayed behind.

Every other answerable item type in the renderer offers a way to clear. The slider was the gap.

## Fix

`SliderField` now renders `ClearInputButton` beside the slider, shown only once the item holds an answer. Placement mirrors `RadioFormGroup`, so the two controls in the reporter's form now look and behave the same way.

The slider and the button sit in a flex `Box` that stacks vertically on narrow viewports, matching the responsive treatment `RadioFormGroup` already uses.

### Why `ClearInputButton` and not `ClearButtonAdornment`

The text-field item types (integer, decimal, string) use `ClearButtonAdornment`, an icon inside the input. It is the wrong fit here:

- It ignores the `hideClearButton` renderer config. `ClearInputButton` honours it.
- It is deliberately `aria-hidden` with `tabIndex={-1}`, because a text input can already be cleared by typing. A slider cannot, so the control needs to be keyboard-reachable.
- Its click handler re-focuses via `.MuiOutlinedInput-root`, which a slider does not have.

`ClearInputButton` also renders the literal `CLEAR` label the reporter was comparing against.

### Clearing removes the answer, it does not write 0

`handleClear` writes an empty QrItem, so the item leaves the `QuestionnaireResponse` entirely. The issue is explicit that `0` is a real answer on these forms and is not the same thing as unanswered.

No change was needed for the cleared visual state: `SliderDisplayValue` already falls back to `-` whenever the item has no answer, and `isInteracted` keys off answer presence rather than the value, so an answer of `0` still reads as answered.

## Changes

- `SliderField.tsx` — new `onClear` prop; renders `ClearInputButton` in a flex wrapper beside the slider. The bulk of the diff is JSX re-indentation from that wrapper.
- `SliderItem.tsx` — 8 lines: `handleClear`, passed to both the repeated and non-repeated call sites.
- `test/clearSliderField.test.tsx` — new, 6 tests built on the `Pupil Size (mm)` item from the issue, following the existing `clearChoiceField.test.tsx` pattern.

No version bump. No public API change: `SliderField` is internal, and `onClear` is a required prop on it only.

## A change I tried and backed out

I also changed the unanswered fallback from `0` to `minValue`, on the theory that a `0` thumb would sit off the left end of the track whenever `minValue > 0`, and fixed a coupled truthiness bug on `SliderItem.tsx:53` (`if (qrItem?.answer[0].valueInteger)` treats a real `0` as absent) that would otherwise have made a `0` answer render at `minValue`.

I probed the DOM before trusting it. With `minValue: 2` and no answer, the output is identical either way — thumb at `left: 0%`, `aria-valuenow=2`. MUI clamps the value into `[min, max]` itself, so the premise was wrong and the change had no observable effect. Reverted rather than carry inert diff, along with the test written to guard it, which passed with the fix stashed and was therefore testing MUI rather than this code.

Line 53's truthiness test is still sloppy and worth tidying, but MUI's clamping makes it harmless today, so it is out of scope here.

## Verification

| Check | Result |
| --- | --- |
| New tests | 6 passed |
| New tests, fix stashed | 4 failed, 2 passed |
| Renderer Jest | 2261 passed, 100 suites |
| `npm run build -w packages/smart-forms-renderer` | clean |
| `npm run lint` | 0 errors, 344 pre-existing warnings |
| Prettier | clean on all changed files |

The 2 tests that pass with the fix stashed are the negative cases — no button while unanswered, none when `readOnly`. They guard against over-rendering, so passing before the change is the correct result.

### DIFF
--- a/packages/smart-forms-renderer/src/components/FormComponents/SliderItem/SliderField.tsx
+++ b/packages/smart-forms-renderer/src/components/FormComponents/SliderItem/SliderField.tsx
@@ -18,12 +18,14 @@
 import React from 'react';
 import type { PropsWithIsTabledAttribute } from '../../../interfaces/renderProps.interface';
 import { getSliderMarks } from '../../../utils/slider';
+import Box from '@mui/material/Box';
 import Stack from '@mui/material/Stack';
 import SliderLabels from './SliderLabels';
 import SliderDisplayValue from './SliderDisplayValue';
 import { useRendererConfigStore } from '../../../stores';
 import { StyledRequiredTypography } from '../Item.styles';
 import { StandardSlider } from './Slider.styles';
+import ClearInputButton from '../ItemParts/ClearInputButton';
 
 interface SliderFieldProps extends PropsWithIsTabledAttribute {
   linkId: string;
@@ -40,6 +42,7 @@ interface SliderFieldProps extends PropsWithIsTabledAttribute {
   readOnly: boolean;
   instructionsId?: string;
   onValueChange: (newValue: number) => void;
+  onClear: () => void;
 }
 
 function SliderField(props: SliderFieldProps) {
@@ -58,7 +61,8 @@ function SliderField(props: SliderFieldProps) {
     readOnly,
     instructionsId,
     isTabled,
-    onValueChange
+    onValueChange,
+    onClear
   } = props;
 
   const readOnlyVisualStyle = useRendererConfigStore.use.readOnlyVisualStyle();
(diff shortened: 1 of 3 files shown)


leoniedickson (commented):
Found a regression while testing this: the slider now renders much narrower than before.

**Repro:** load any slider item (e.g. the `SliderInteger` story in Storybook) and compare against `main`. On `main` the slider fills out to `textFieldWidth` (320px by default). On this branch it collapses to exactly 160px — the `minWidth` floor in `sliderSx`.

**Root cause:** the new wrapper `Box` around the slider + `ClearInputButton` (`SliderField.tsx:83`) makes the slider's `Stack` a flex item instead of a plain block child. Flex items default to shrink-to-content sizing rather than stretching, so:
- On `sm+` (row direction), the `Stack` has no `flexGrow`/`width`, so it sizes to content and shrinks toward `minWidth: 160` instead of filling out to `maxWidth: 320`.
- On `xs` (column direction), `alignItems: 'start'` disables the cross-axis stretch that previously gave the `Stack` full width for free (it was a plain block element before, so `width: auto` filled the container up to `maxWidth`).

I verified this empirically: swapping in `main`'s version of `SliderField.tsx` and screenshotting the `SliderInteger` story gives a 320px-wide slider; this branch's version gives 160px.

**Suggested fix**, tested against both breakpoints (desktop row layout and mobile column layout, screenshotted via Storybook):

```diff
         display="flex"
         alignItems={{ xs: 'start', sm: 'center' }}
         flexDirection={{ xs: 'column', sm: 'row' }}
-        columnGap={1}>
-        <Stack sx={{ ...sliderSx }}>
+        gap={1}>
+        <Stack sx={{ ...sliderSx, width: '100%', flexGrow: 1 }}>
```

(Bundled in the same line: `columnGap` → `gap`, since `column-gap` has no effect once `flexDirection` switches to `'column'` on mobile — same underlying "the new flex wrapper needs more care" issue, just cosmetic rather than a width collapse.)

With this change the slider is back to 320px on desktop with an 8px gap before `CLEAR`, and stacks full-width above `CLEAR` on mobile. All 6 tests in `clearSliderField.test.tsx` still pass.

clinnygee:
Fixed in 0b3f7619 — applied your suggested diff as-is (`gap` and `width: 100%` / `flexGrow: 1` on the `Stack`). Slider is back to 320px on desktop and full-width on mobile. Thanks for catching it.

PR #2111, opened 2026-09-08
fix(#2107): merge multi-value template-extract splits into one array element

### LINKED ISSUE #2107: sdc-template-extract: repeating value split into separate array elements, in reverse order, when an earlier sibling value is empty
# sdc-template-extract: repeating value split into separate array elements, in reverse order, when an earlier sibling value is empty

**Describe the bug**

Within a single `templateExtractContext`, if the *first*-declared `templateExtractValue` evaluates to an empty result (e.g. an unanswered question) and a *later*-declared `templateExtractValue` evaluates to **multiple** values (a repeating question), the multiple values are written as separate array elements instead of merged into one — and in reverse order.

Root cause is the interaction of two things in `packages/sdc-template-extract/src/SDCExtractQuestionnaireResponseOperation/utils`:

1. `evaluateAndInsertIntoPath` (`evaluateTemplateExtractPath.ts`) sets `isNewInsert = true` per context match and only flips it to `false` once some `valuePathMap` entry evaluates non-empty:
   ```ts
   let isNewInsert = true;
   for (const [valuePath, valueEvaluation] of valuePathMap.entries()) {
     const valueResult = fhirPathEvaluate(...);
     insertValuesToPath(entryPath, insertIndex, isNewInsert, valuePath, valueResult, ...);
     if (Array.isArray(valueResult) && valueResult.length) {
       isNewInsert = false;
     }
   }
   ```
   `valuePathMap` is walked in template-declaration order. If the first-declared value (e.g. `_family`) has no answer, it evaluates to `[]`, so `isNewInsert` is still `true` when the next value (e.g. `_given`) runs.

2. `insertValuesToPath` (`templateInsert.ts`) reuses that *same* `isNewInsert` for every item produced by a single value's multi-value evaluation:
   ```ts
   for (let i = 0; i < valuesToInsert.length; i++) {
     walkTemplateAndInsertValue(templateToMutate, entryPath, cleanedEntryPathSegments, isNewInsert, valuesToInsert[i]);
   }
   ```
   For a repeating question, `buildValuesToInsert` splits the evaluated result into one object per answer (e.g. `[{given:['a']}, {given:['b']}]`) — intentionally, so the first seeds the array element and the rest merge into it. But since `isNewInsert` is still `true` for *both* items here, **both** hit the "new array insert" branch in `walkTemplateAndInsertValue`:
   ```ts
   if (Array.isArray(current) && typeof finalSegment === 'number' && isNewInsert) {
     current.splice(finalSegment, 0, valueToInsert);
     return;
   }
   ```
   at the *same* `insertIndex` — the second splice pushes the first element over instead of merging into it.

**To Reproduce**

Minimal `Patient.name[0]` template with a `templateExtractContext`, a `_family` value (declared first) and a repeating `_given` value (declared second). Leave `family` unanswered; answer `given` twice.

Questionnaire (contained template + item definitions):
```json
{
  "id": "GivenNameReversal",
  "url": "https://smartforms.csiro.au/docs/tests/GivenNameReversal",
  "status": "draft",
  "subjectType": ["Patient"],
  "contained": [
    {
      "resourceType": "Patient",
      "id": "patTemplate",
      "name": [
        {
          "extension": [
            { "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractContext", "valueString": "item.where(linkId = 'name')" }
          ],
          "_family": {
            "extension": [
              { "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractValue", "valueString": "item.where(linkId = 'family').answer.value.first()" }
            ]
          },
          "_given": [
            {
              "extension": [
                { "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractValue", "valueString": "item.where(linkId = 'given').answer.value" }
              ]
            }
          ]
        }
      ]
    }
  ],
  "item": [
    {
      "extension": [
        { "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtract",
          "extension": [{ "url": "template", "valueReference": { "reference": "#patTemplate" } }] }
      ],
      "linkId": "patient", "text": "Patient Information", "type": "group",
      "item": [
        { "linkId": "name", "text": "Name", "type": "group", "repeats": true,
          "item": [
            { "linkId": "given", "text": "Given Name(s)", "type": "string", "repeats": true },
            { "linkId": "family", "text": "Family/Surname", "type": "string" }
          ]
        }
      ]
    }
  ]
}
```

QuestionnaireResponse (only `given` answered, twice; `family` left blank):
```json
{
  "resourceType": "QuestionnaireResponse",
  "status": "completed",
  "questionnaire": "https://smartforms.csiro.au/docs/tests/GivenNameReversal",
  "item": [
    { "linkId": "patient", "item": [
      { "linkId": "name", "item": [
        { "linkId": "given", "answer": [{ "valueString": "a" }, { "valueString": "b" }] }
      ]}
    ]}
  ]
}
```

Run `$extract`.

**Expected behavior**

```json
"name": [{ "given": ["a", "b"] }]
```

**Actual**

```json
"name": [
  { "given": ["b"] },
  { "given": ["a"] }
]
```

Two separate `name` entries instead of one, and the values are in reverse order.

**Additional context**

- This only surfaces when the *first*-declared `valuePathMap` entry for a context evaluates empty while a *later*-declared one evaluates to multiple values. If `family` is answered (as in the existing `QComplexTemplateExtract` test fixture), `isNewInsert` flips to `false` before `_given` runs, the merge branch is used instead, and the existing tests pass — which is why this wasn't previously caught.
- Confirmed via `git diff` that this is unrelated to #2093 (the `templateExtractValue`-into-existing-array-element fix) — the `isNewInsert`/splice code path involved here is untouched by that PR.
**Proposed fix** (verified against the repro above plus the full existing test suite — 70/70 passing, no regressions):

```ts
// insertValuesToPath, templateInsert.ts
let isFirstInsert = isNewInsert;
for (let i = 0; i < valuesToInsert.length; i++) {
  const valueToInsert = valuesToInsert[i];
  const cleanedEntryPathSegments = cleanEntryPathSegments(entryPathSegments, insertIndex);

  walkTemplateAndInsertValue(
    templateToMutate,
    entryPath,
    cleanedEntryPathSegments,
    isFirstInsert,
    valueToInsert
  );
  isFirstInsert = false;
}
```

`isNewInsert` (as passed in) answers "is this the first value for this context result overall" — correct for item 0 of a multi-value split. But items 1..N of that same split are, by construction, going into the array element item 0 just created (or that already existed) — they should always merge, regardless of what the outer per-context flag says. The current code conflates "first value for this context" with "first value for this particular splice," and those only coincide when a value never splits into more than one item. Tracking a local flag that's forced to `false` after the loop's first iteration fixes it without touching the outer `isNewInsert` logic at all.

Verified with the repro above (`given: ["a","b"]`, correctly ordered, single element) and a 3-value case (`given: ["a","b","c"]`).

leoniedickson:
## Summary
- Fixes #2107: within a single `templateExtractContext`, when an earlier-declared `templateExtractValue` evaluates empty and a later-declared one evaluates to multiple values (e.g. a repeating question), the values were being split into separate array elements in reverse order instead of merged into one.
- Root cause: `insertValuesToPath` reused the caller's `isNewInsert` flag for every item produced by a single value's multi-value split, instead of only the first. Introduces a local `isFirstInsert` flag so only the first item of a split can create a new array element; subsequent items from that same split always merge into it, regardless of the outer flag.

## Test plan
- [x] Added `templateExtractGivenNameReversal.test.ts` reproducing the issue's exact repro (unanswered `family`, repeating `given`), plus a 3-value variant and a case with the repeating value declared *before* the single value (both answered) — the latter shows the bug isn't actually about declaration order.
- [x] Verified each new test fails against the pre-fix code with the exact symptom described (reversed order, split elements) and passes with the fix.
- [x] Full `sdc-template-extract` suite passes: 72/72, no regressions (including the `QComplexTemplateExtract` real-world fixture).

### DIFF
--- a/packages/sdc-template-extract/src/SDCExtractQuestionnaireResponseOperation/tests/templateExtractGivenNameReversal.test.ts
+++ b/packages/sdc-template-extract/src/SDCExtractQuestionnaireResponseOperation/tests/templateExtractGivenNameReversal.test.ts
@@ -1,0 +1,210 @@
+import { extract } from '../utils';
+import type { OutputParameters, ReturnParameter } from '../interfaces';
+import { createInputParameters } from '../utils/createInputParameters';
+import type { Bundle, Patient, Questionnaire, QuestionnaireResponse } from 'fhir/r4';
+
+// Mock the fetchQuestionnaire callback function
+const mockFetchQuestionnaire = jest.fn();
+const mockFetchQuestionnaireConfig = {
+  sourceServerUrl: 'https://example.com/fhir',
+  headers: { Authorization: 'Bearer token' }
+};
+
+// Regression test for https://github.com/aehrc/smart-forms/issues/2107
+// Within a single templateExtractContext, a first-declared templateExtractValue evaluating to
+// an empty result (unanswered "family") must not cause a later-declared repeating
+// templateExtractValue ("given") to be split into separate, reversed-order array elements.
+const QGivenNameReversal: Questionnaire = {
+  resourceType: 'Questionnaire',
+  id: 'GivenNameReversal',
+  url: 'https://smartforms.csiro.au/docs/tests/GivenNameReversal',
+  status: 'draft',
+  subjectType: ['Patient'],
+  contained: [
+    {
+      resourceType: 'Patient',
+      id: 'patTemplate',
+      name: [
+        {
+          extension: [
+            {
+              url: 'http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractContext',
+              valueString: "item.where(linkId = 'name')"
+            }
+          ],
+          _family: {
+            extension: [
+              {
+                url: 'http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractValue',
+                valueString: "item.where(linkId = 'family').answer.value.first()"
+              }
+            ]
+          },
+          _given: [
+            {
+              extension: [
+                {
+                  url: 'http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractValue',
+                  valueString: "item.where(linkId = 'given').answer.value"
+                }
+              ]
+            }
+          ]
+        } as any
+      ]
+    }
+  ],
+  item: [
+    {
+      extension: [
+        {
+          url: 'http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtract',
+          extension: [
+            {
+              url: 'template',
+              valueReference: { reference: '#patTemplate' }
+            }
+          ]
+        }
+      ],
+      linkId: 'patient',
+      text: 'Patient Information',
+      type: 'group',
+      item: [
+        {
+          linkId: 'name',
+          text: 'Name',
+          type: 'group',
+          repeats: true,
+          item: [
+            { linkId: 'given', text: 'Given Name(s)', type: 'string', repeats: true },
+            { linkId: 'family', text: 'Family/Surname', type: 'string' }
+          ]
+        }
+      ]
+    }
+  ]
+};
+
+function buildQR(givenAnswers: string[], familyAnswer?: string): QuestionnaireResponse {
+  return {
+    resourceType: 'QuestionnaireResponse',
+    status: 'completed',
+    questionnaire: 'https://smartforms.csiro.au/docs/tests/GivenNameReversal',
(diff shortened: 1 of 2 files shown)


clinnygee (approved):
lgtm

PR #2114, opened 2026-09-09
fix(#2112): viewer's Save as Final now writes back extracted data

### LINKED ISSUE #2112: Save as Final on the response preview screen (/viewer) does not perform writeback
**Describe the bug**
When opening an existing, previously-saved `QuestionnaireResponse` from the dashboard (Responses → Open Response → preview screen), the `Save as Final` button only saves the `QuestionnaireResponse` — it does not perform template-based or observation-based extraction, and does not write the extracted data back to the FHIR server. This is inconsistent with the `Save as Final` action available on the renderer (`/renderer` and `/renderer/preview`), which extracts and writes back data for questionnaires that support it.

**To Reproduce**
1. Complete and save a response to a questionnaire that supports extraction/writeback (e.g. one with a template-based or observation-based extraction extension).
2. Go to the dashboard → Responses → open that saved response (routes to `/viewer`).
3. Click `Save as Final` on the preview screen and confirm.
4. Check the FHIR server: no extracted resources (e.g. Observations) were written back, even though the same questionnaire triggers writeback when saved as final from `/renderer`.

**Expected behavior**
`Save as Final` on the `/viewer` preview screen should behave the same as `Save as Final` on the renderer: detect the extract mechanism (`getExtractMechanism`), run extraction, and prompt/write back the extracted bundle to the FHIR server before/alongside saving the `QuestionnaireResponse`.

**Root cause (from investigation)**
`ViewerSaveAsFinal.tsx` / `ViewerSaveAsFinalDialog.tsx` (`apps/smart-forms-app/src/features/viewer/ViewerNav/SaveAsFinal/`) is a separate, older implementation that predates the extraction/writeback feature added to `SaveAsFinalAction.tsx` (`apps/smart-forms-app/src/features/renderer/components/RendererActions/`). The viewer's `handleSave()` calls `saveQuestionnaireResponse` directly with no extraction step, while the renderer's flow runs `handleTemplateExtract`/`handleObservationExtract` and then `RendererSaveAsFinalWriteBackDialog.handleWriteBack()`, which POSTs the extracted bundle. The viewer also uses `saveQuestionnaireResponse` instead of `saveProgress` (minor secondary inconsistency — `saveProgress` additionally strips empty answers).

**Suggested fix**
Extract the extraction/writeback logic out of `SaveAsFinalAction.tsx` into a shared hook, and reuse it in `ViewerSaveAsFinal`/`ViewerSaveAsFinalDialog` so both entry points share one save+writeback code path instead of two divergent implementations.

leoniedickson:
## Summary
- Fixes #2112: the `/viewer` preview screen's "Save as Final" button used a separate, older save path that never got the extraction/writeback feature the renderer's "Save as Final" has. Both entry points now share a `useSaveAsFinalExtraction` hook and the `SaveAsFinalOnlyDialog`/`SaveAsFinalWriteBackDialog` components, moved into `features/writeBack` since they're no longer renderer-only.
- Fixes a silent-failure gap surfaced by review: template-based extraction could bail out with only a `console.error` (or nothing) when `patient`/`user` hadn't resolved or extraction returned an `OperationOutcome`, leaving the button revert with no feedback — newly reachable from the viewer, which previously had no extraction step at all. Both bail paths now show an error snackbar.

## Test plan
- [x] `tsc --noEmit` passes
- [x] `eslint` passes (no new warnings/errors)
- [x] Full Jest suite passes (47 suites / 647 tests), including new unit tests for `useSaveAsFinalExtraction` covering template-based/observation-based extraction, bundle validation opt-in, the missing-context bail-out, the `OperationOutcome` failure path, and the new error snackbar
- [ ] Manual click-through in the browser: open an in-progress response with a template/observation-extractable questionnaire from `/dashboard/responses`, hit Save as Final from the preview screen, confirm the writeback dialog appears and the bundle POSTs


clinnygee (approved):
lgtm

